PR #1, opened 2020-02-17
Fetch buildroot sources from lookaside cache

iliana:
A future PR will switch everything else over to `sdk-fetch`, but want to get this reviewed first.

Tested by running `make` and the result of that is the image I've been using for builds yesterday/today.


tjkirch (commented):
Would you please add the background and your testing done to the description?

This seems like an approach we could use for our other Dockerfiles too.

jamieand (approved):

tjkirch (approved):

Issue #25, opened 2020-04-23
Update Rust to 1.43.1


tjkirch:
https://blog.rust-lang.org/2020/04/23/Rust-1.43.0.html

tjkirch:
1.43.1 is now out.  https://blog.rust-lang.org/2020/05/07/Rust.1.43.1.html

Issue #24, opened 2020-04-07
Update buildroot


tjkirch:
We're currently on 2019.08.3 and 2020.02 is available.

bcressey:
As part of this I'd like to also ensure we move to 5.4 kernel headers, since that's now our minimum expected kernel version.

That will let us take advantage of some new features, such as systemd's support for pidfd.

PR #27, opened 2020-05-14
update SDK dependencies

bcressey:
**Issue number:**
#24 #25 #26

**Description of changes:**
Update base layer to Fedora 32, which provides GCC 10.1.1 and binutils 2.34 for the host toolchain.

Move to Buildroot 2020.02.2, which gets us GCC 9.3.0, binutils 2.33.1, and kernel 5.4 headers. The patch to add GCC 9.2.0 is no longer needed.

Update to Rust 1.43.1 and Go 1.14.2. Also update musl, libunwind, and license data for completeness.

**Testing done:**
Built the k8s 1.15 variant, which needed some minor fixes. Made an AMI, joined to cluster, ran conformance tests.

```
[    0.000000] Linux version 5.4.38 (builder@buildkitsandbox) (gcc version 9.3.0 (Buildroot 2020.02.2)) #1 SMP Thu May 14 00:04:49 UTC 2020
```

**Terms of contribution:**


iliana (approved):

tjkirch (approved):
> Built the k8s 1.15 variant, which needed some minor fixes

I didn't see any specific fixes when going through the changes, but if I missed them, it seems like a good idea to check 1.16 too.

bcressey:
> I didn't see any specific fixes when going through the changes, but if I missed them, it seems like a good idea to check 1.16 too.

There wasn't anything specific to 1.15. I needed to fix the systemd build to work with the newer version of meson, and the selinux policy to work with the newer CIL compiler. I'll put the details in the Bottlerocket PR.

PR #30, opened 2020-08-07
Bump all the things

### LINKED ISSUE #29: Update Rust to 1.45.2
Some unsoundness fixes:
* https://blog.rust-lang.org/2020/07/16/Rust-1.45.0.html
* https://blog.rust-lang.org/2020/07/30/Rust-1.45.1.html
* https://blog.rust-lang.org/2020/08/03/Rust-1.45.2.html

iliana:
**Issue number:**

n/a

**Description of changes:**

- glibc 2.31 -> 2.32
- musl 1.2.0 -> 1.2.1
- Rust 1.43.1 -> 1.45.2
  * Modified Rust LLVM library stripping because `libLLVM-*.so` is no longer part of `rustlib`; I believe this is related to [the new on-by-default `embed-bitcode` compiler flag in Rust 1.45](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1450-2020-07-16)
- Go 1.14.2 -> 1.14.7
- Update license-scan deps

**Testing done:**

Successfully builds Bottlerocket with some changes, which will go up on a draft PR on the OS repo.

**Terms of contribution:**


etungsten (approved):

tjkirch (approved):
LGTM, assuming we can get the glibc symbol versioning / host/target issue figured out with @bcressey.

bcressey:
The glibc bump here might be OK - the problems we'd see would be that the Go + Rust toolchains would have missing symbols, which would show up pretty quickly when building the OS.

bcressey (approved):

bcressey:
Just saw the note in https://github.com/bottlerocket-os/bottlerocket/pull/1014 - so let's back out the glibc update until I can figure out the best way to fix.

PR #33, opened 2020-08-30
update SDK to v0.13.0

### LINKED ISSUE #28: [Feature Request] Support building on aarch64 hosts (M6g, etc.)
**Image I'm using:** 
bottlerocket-aws-k8s-1.15-aarch64.img

**Issue or Feature Request:**
Feature Request: support building on aarch64 hosts. 
Attempting to build on an aarch64 host results in a number of "exec format error"s, most likely from docker pulling x86-based images. Using `cargo make -e BUILDSYS_ARCH=aarch64` I was able to build an aarch64 img and amiize it from an x86_64 host.  
It would be nice to be able build natively on aarch64, especially as part of larger EKS+Bottlerocket on Arm toolchain. 

Thanks!

### LINKED ISSUE #31: glibc needs to be the same version across the Fedora base image, the SDK, and the OS
Some of our build scripts are built with the Bottlerocket toolchain, but run with the Fedora glibc version in the SDK image. If we try to upgrade glibc in the OS ahead of the SDK, we get symbol version mismatches; if we try to upgrade glibc in the SDK ahead of the Fedora base image, we get symbol version mismatches.

This issue is to track unwinding the dependency of things built with the Bottlerocket toolchain on the particulars of the Fedora glibc.

### LINKED ISSUE #32: update Rust to 1.46
[Rust 1.46 is out!](https://blog.rust-lang.org/2020/08/27/Rust-1.46.0.html)

bcressey:
**Issue number:**
Closes #28 
Closes #31
Closes #32 

**Description of changes:**
The major change is the additional of new, vendor-specific Rust targets. These are copied from the "unknown" vendor targets we used previously. The goal is to eliminate the host/target confusion for Rust that existed when building for `x86_64-unknown-linux-gnu` on x86_64, or for `aarch64-unknown-linux-gnu` on aarch64.

This caused a few subtle issues, such as linking host programs using our target GCC toolchain and C library, which prevented us from updating glibc either here or in the OS. Now we're free to take the glibc update.

The update to Rust 1.46.0 also includes aarch64 versions of the stage-0 artifacts, so that we can build the SDK on aarch64 hosts. Enabling this is another happy side effect of the new vendor targets.

**Testing done:**
Built the four supported SDK combinations:
* x86_64 host, x86_64 target
* x86_64 host, aarch64 target
* aarch64 host, x86_64 target
* aarch64 host, aarch64 target

Verified that an x86_64 host could build the OS for both targets, and likewise that an aarch64 host could build the OS for both targets. Verified that the resulting artifacts worked as expected when registered as AMIs.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -277,7 +277,7 @@ RUN \
 
 ARG ARCH
 ARG VENDOR="bottlerocket"
-ARG RUSTVER="1.45.2"
+ARG RUSTVER="1.46.0"
 
 USER builder
 WORKDIR /home/builder
--- a/Dockerfile
+++ b/Dockerfile
@@ -276,7 +276,7 @@ RUN \
   chown -R builder:builder /usr/libexec/rust
 
 ARG ARCH
-ARG TARGET="${ARCH}-bottlerocket-linux-gnu"
+ARG VENDOR="bottlerocket"
 ARG RUSTVER="1.45.2"
 
 USER builder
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -292,6 +292,19 @@ WORKDIR /home/builder/rust
...
+  for libc in gnu musl ; do \
+    cp src/librustc_target/spec/${ARCH}_{unknown,${VENDOR}}_linux_${libc}.rs ; \
+    sed -i -e 's@target_vendor: "unknown"@target_vendor: "'${VENDOR}'"@' \
+      src/librustc_target/spec/${ARCH}_${VENDOR}_linux_${libc}.rs ; \
+    sed -i -e '/("'${ARCH}-unknown-linux-${libc}'", .*),/a("'${ARCH}-${VENDOR}-linux-${libc}'", '${ARCH}_${VENDOR}_linux_${libc}'),' \
+      src/librustc_target/spec/mod.rs ; \
+  done && \
+  grep -Fq ${VENDOR} src/librustc_target/spec/mod.rs && \
+  grep -Fq ${VENDOR} src/librustc_target/spec/${ARCH}_${VENDOR}_linux_gnu.rs && \
+  grep -Fq ${VENDOR} src/librustc_target/spec/${ARCH}_${VENDOR}_linux_musl.rs
```
tjkirch, line 333: I like these safety checks.  Would it be possible to check that there aren't any `unknown-linux` references remaining in related files, too?  To catch any future changes in the files that would require another sed.

tjkirch (approved):
Ben the personal trainer be like
```
 _____________________ 
< Never skip ARM day! >
 --------------------- 
   \
    \
        .--.
       |o_o |
       |:_/ |
      //   \ \
     (|     | )
    /'\_   _/`\
    \___)=(___/
```

Dockerfile
bcressey: `mod.rs` will include the existing `unknown-linux` entries, which we want to keep. The target-specific files will also have them for `llvm_target`, e.g. `llvm_target: "aarch64-unknown-linux-musl".to_string()`.

I am concerned about future changes too, just not quite sure if there's a good way to catch them.

iliana (approved):

Dockerfile
```diff
@@ -292,6 +292,19 @@ WORKDIR /home/builder/rust
...
   dir=build/cache/$(awk '/^date:/ { print $2 }' src/stage0.txt); \
   mkdir -p $dir && mv ../*.xz $dir
+
+RUN \
+  for libc in gnu musl ; do \
+    cp src/librustc_target/spec/${ARCH}_{unknown,${VENDOR}}_linux_${libc}.rs ; \
+    sed -i -e 's@target_vendor: "unknown"@target_vendor: "'${VENDOR}'"@' \
+      src/librustc_target/spec/${ARCH}_${VENDOR}_linux_${libc}.rs ; \
+    sed -i -e '/("'${ARCH}-unknown-linux-${libc}'", .*),/a("'${ARCH}-${VENDOR}-linux-${libc}'", '${ARCH}_${VENDOR}_linux_${libc}'),' \
+      src/librustc_target/spec/mod.rs ; \
```
webern, line 329: Maybe worthy of an explanation for the less gifted?

webern (approved):

Dockerfile
bcressey: I've added an explanation; let me know if it suffices or if I should expand on the details.

What this amounts to is the cheapest way I could find to copy existing Rust targets, change the "vendor" field, and integrate them into `rustc` so we could use them.

tjkirch (approved):

Dockerfile
webern: That's great! 🎓

PR #37, opened 2020-11-06
update toolchain and add archive target

### LINKED ISSUE #7: create toolchain archive from SDK
For best results we need out-of-tree kernel modules to be compiled with the same toolchain that we use to build Bottlerocket.

However, since those builds will happen inside a container, it will be awkward to use the SDK if it is also a container. You'd need to install Docker or use `ctr` to interact with `containerd.sock` on the host, and mount in the same `/usr/src/kernels` and `/lib/modules` paths that are already mounted into host containers.

The full SDK is also rather large and we only need a small part of it - GCC and binutils - to build kernel modules.

A better approach would be to extract the toolchain into a separate archive. That would play nicely with other tools and automation like DKMS.

We'll need to publish the archives somewhere, and also figure out how to match up the host OS with the SDK version so the correct version can be installed.

### LINKED ISSUE #34: update Go to 1.15
[Go 1.15 is out!](https://golang.org/doc/go1.15)

### LINKED ISSUE #36: Update to Rust 1.47
https://blog.rust-lang.org/2020/10/08/Rust-1.47.html
https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1470-2020-10-08
:rocket:

bcressey:
**Issue number:**
Partly addresses #7, fixes #34 and fixes #36.

**Description of changes:**
Create a toolchain archive target that we can use to extract gcc and binutils from the SDK, so that it can be used in contexts where running a container is difficult.

To make the toolchain work on distros other than a recent Fedora, build the toolchain with an older Ubuntu.

Update Go to 1.15.4, Rust to 1.47.

**Testing done:**
Verified that I can use the SDK to build Bottlerocket on both architectures (x86_64, aarch64) for both architectures. A small fix is needed for Go programs to add `linkmode=external` to ldflags, since the default linker for `buildmode=pie` changed in Go 1.15.

Confirmed that the toolchain archive has the expected contents:
```
❯ tar tf bottlerocket-toolchain-aarch64-v0.14.0.aarch64.tar.xz
...
toolchain/usr/bin/aarch64-bottlerocket-linux-musl-gcc
...
toolchain/licenses/gcc/COPYING3
toolchain/licenses/gcc/COPYING.RUNTIME
toolchain/licenses/gcc/COPYING3.LIB
toolchain/licenses/gcc/COPYING.LIB
toolchain/licenses/gcc/COPYING
...
```

Verified that I can use the toolchain archive to build out-of-tree kernel modules:
* x86_64 - Falco, Wireguard, ZFS
* aarch64 - Wireguard, ZFS

Tested the module build using Amazon Linux 2 and Fedora 32 base layers on both architectures.

Confirmed that the module could be loaded on instances for both architectures.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,21 @@ COPY ./sdk-fetch /usr/local/bin
 
 # =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
-FROM base as toolchain
+# We expect our C cross-compiler to be used on other distros for building kernel
+# modules, so we build it with an older glibc for compatibility.
+FROM ubuntu:16.04 as compat
+RUN \
+  apt-get update && \
+  apt-get -y dist-upgrade && \
+  apt-get -y install \
+    autoconf automake bc build-essential cpio curl file git \
+    libexpat1-dev libtool libz-dev pkgconf python3 unzip wget && \
+  useradd -m -u 1000 builder
+COPY ./sdk-fetch /usr/local/bin
+
+# =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+FROM compat as toolchain
 USER builder
 
 # Configure Git for any subsequent use.
@@ -57,6 +71,7 @@ RUN \
   find output/${ARCH}-gnu/build/linux-headers-${KVER}/usr/include -name '.*' -delete
 
 WORKDIR /home/builder/buildroot/output/${ARCH}-gnu/build
+SHELL ["/bin/bash", "-c"]
 RUN \
   install -p -m 0644 -Dt licenses/binutils host-binutils-*/COPYING{,3}{,.LIB} && \
   install -p -m 0644 -Dt licenses/bison host-bison-*/COPYING && \
@@ -67,8 +82,7 @@ RUN \
   install -p -m 0644 -Dt licenses/linux linux-headers-*/{COPYING,LICENSES/preferred/GPL-2.0,LICENSES/exceptions/Linux-syscall-note} && \
   install -p -m 0644 -Dt licenses/m4 host-m4-*/COPYING && \
   install -p -m 0644 -Dt licenses/mpc host-mpc-*/COPYING.LESSER && \
-  install -p -m 0644 -Dt licenses/mpfr host-mpfr-*/COPYING{,.LESSER} && \
-  install -p -m 0644 -Dt licenses/tar host-tar-*/COPYING
+  install -p -m 0644 -Dt licenses/mpfr host-mpfr-*/COPYING{,.LESSER}
 
 # =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -349,7 +372,7 @@ FROM sdk-libc as sdk-go
 
 ARG ARCH
 ARG TARGET="${ARCH}-bottlerocket-linux-gnu"
-ARG GOVER="1.14.7"
+ARG GOVER="1.15.3"
```
tjkirch:
```suggestion
ARG GOVER="1.15.4"
```
:(

hashes/buildroot
```diff
@@ -26,5 +26,3 @@ SHA512 (mpc-1.1.0.tar.gz) = 72d657958b07c7812dc9c7cbae093118ce0e454c68a585bfb0e2
 SHA512 (mpfr-4.0.2.tar.xz) = d583555d08863bf36c89b289ae26bae353d9a31f08ee3894520992d2c26e5683c4c9c193d7ad139632f71c0a476d85ea76182702a98bf08dde7b6f65a54f8b88
 # http://cvsweb.netbsd.org/bsdweb.cgi/src/sys/sys/queue.h?rev=1.70
 SHA512 (queue.h) = 2f0d5e6e4dc3350285cf17009265dddcbe12431c111868eea39bc8cb038ab7c1f2acacbb21735c4e9d4a1fd106a8fc0f8611ea33987d4faba37dde5ce6da0750
-# https://mirrors.kernel.org/gnu/tar/tar-1.29.cpio.gz
```
tjkirch, line 29: Was this unused?

Makefile
```diff
@@ -2,17 +2,36 @@ ARCH ?= $(shell uname -m)
...
+
+$(TOOLCHAIN_ARCHIVE) :
+	@DOCKER_BUILDKIT=1 docker build . \
+		--tag $(TOOLCHAIN_TAG) \
+		--target toolchain-final \
+		--squash \
+		--build-arg ARCH=$(ARCH)
+	@docker run --rm --entrypoint cat $(TOOLCHAIN_TAG) /tmp/toolchain.tar.xz > $(@)
+
+upload : $(SDK_ARCHIVE) $(TOOLCHAIN_ARCHIVE)
```
tjkirch, line 30: Seems like we might want a `download-toolchain` target now too?

tjkirch (approved):

Dockerfile
bcressey: 😢

hashes/buildroot
bcressey: Seems like it's not needed on Ubuntu? Not entirely sure why.

Makefile
bcressey: We'll handle the download and conversion into the "kmod kit" in Bottlerocket, since that's where we bring together the kernel and the toolchain.

bcressey:
I'm not expecting any problems, but I've converted this back to draft until I can re-test the distro build after the Go change from 1.15.3 to 1.15.4.

tjkirch (approved):

Dockerfile
```diff
@@ -57,6 +71,7 @@ RUN \
   find output/${ARCH}-gnu/build/linux-headers-${KVER}/usr/include -name '.*' -delete
 
 WORKDIR /home/builder/buildroot/output/${ARCH}-gnu/build
+SHELL ["/bin/bash", "-c"]
```
iliana, line 74: Can we set this at the beginning of `FROM ubuntu:16.04 as compat` so we don't have to set it in multiple places?

iliana (approved):

ajorg-aws (approved):
No open source concerns with switching to `ubuntu:16.04` for this.

Dockerfile
bcressey: I have a slight preference to keep it here, near the place where we rely on brace expansion.

PR #54, opened 2021-06-08
add govc

jpculp:
**Issue number:**

https://github.com/bottlerocket-os/bottlerocket/issues/1605

**Description of changes:**

Add `govmomi/govc`, which provides the host tool needed to support uploading OVAs.

**Testing done:**

Built SDK with `govc` on both architectures. Also tested cross-compilation.

```
$ /usr/libexec/tools/govc version -l
Build Version: 0.26.0
Build Commit: 34586b6
Build Date: 2021-06-08T18:43:48Z
```

Verified license artifacts for `govmomi` and its dependencies are present in `/usr/share/licenses/govmomi`.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -494,6 +494,45 @@ RUN \
 
 # =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
+FROM sdk-go as sdk-govc
+
+USER root
+RUN \
+  mkdir -p /usr/libexec/tools /usr/share/licenses/govmomi && \
+  chown -R builder:builder /usr/libexec/tools /usr/share/licenses/govmomi
+
+ARG GOVMOMIVER="0.26.0"
+ARG GOVMOMISHORTCOMMIT="34586b6"
+ARG GOVMOMIDATE="2021-06-03T19:03:25Z"
+
+USER builder
+WORKDIR ${GOPATH}/src/github.com/vmware/govmomi
+COPY ./hashes/govmomi /home/builder/hashes
+RUN \
+  sdk-fetch /home/builder/hashes && \
+  tar --strip-components=1 -xf govmomi-${GOVMOMIVER}.tar.gz && \
+  rm govmomi-${GOVMOMIVER}.tar.gz
+
+COPY --chown=0:0 --from=sdk-license-scan /usr/libexec/tools/ /usr/libexec/tools/
+RUN \
+  cp -p LICENSE.txt /usr/share/licenses/govmomi && \
+  go mod vendor && \
+  /usr/libexec/tools/bottlerocket-license-scan \
+    --spdx-data /usr/libexec/tools/spdx-data \
+    --out-dir /usr/share/licenses/govmomi \
+    go-vendor ./vendor
+
+RUN \
+  export CGO_ENABLED=0 ; \
+  export BUILD_VERSION_PKG="github.com/vmware/govmomi/govc/flags" ; \
+  go build -mod=vendor -o /usr/libexec/tools/govc -ldflags " \
+    -X ${BUILD_VERSION_PKG}.BuildVersion=${GOVMOMIVER} \
+    -X ${BUILD_VERSION_PKG}.BuildCommit=${GOVMOMISHORTCOMMIT} \
+    -X ${BUILD_VERSION_PKG}.BuildDate=${GOVMOMIDATE} \
+    " github.com/vmware/govmomi/govc
+
+# =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
 FROM sdk as toolchain-archive
 
 ARG ARCH
@@ -565,6 +604,10 @@ COPY --chown=0:0 --from=sdk-license-scan /usr/share/licenses/bottlerocket-licens
 # "sdk-cargo-deny" has the cargo deny command and licenses.
 COPY --chown=0:0 --from=sdk-cargo-deny /usr/libexec/tools/ /usr/libexec/tools/
 COPY --chown=0:0 --from=sdk-cargo-deny /usr/share/licenses/cargo-deny/ /usr/share/licenses/cargo-deny/
+
+# "sdk-govc" has the VMware govc tool and licenses.
+COPY --chown=0:0 --from=sdk-govc /usr/libexec/tools/govc /usr/libexec/tools/govc
+COPY --chown=0:0 --from=sdk-govc /usr/share/licenses/govmomi /usr/share/licenses/govmomi
 
 # Add Rust programs and libraries to the path.
 # Also add symlinks to help out with sysroot discovery.
--- a/hashes/govmomi
+++ b/hashes/govmomi
@@ -1,0 +1,2 @@
+# https://github.com/vmware/govmomi/archive/v0.26.0.tar.gz#/govmomi-0.26.0.tar.gz
+SHA512 (govmomi-0.26.0.tar.gz) = 51e1e41fd230781b841ed238ce87d81e6b6902c67f1aae91c1245bcb69a696d13baaadae4f67e59bf58b018c6fc1e808168eb55eb4b351f82ace8a16633daa5c


Dockerfile
```diff
@@ -495,6 +495,64 @@ RUN \
...
+  mkdir -p ${GOPATH}/src/github.com/vmware && \
+  mv govmomi-${GOVMOMIVER} ${GOPATH}/src/github.com/vmware/govmomi && \
+  ln -s ${GOPATH}/src/github.com/vmware/govmomi
+
+ARG GOOS="linux"
+ARG CGO_ENABLED=0
+ARG PKGPATH="github.com/vmware/govmomi/govc/flags"
+ARG GOARCH_aarch64="arm64"
+ARG GOARCH_x86_64="amd64"
+ARG GOARCH_ARCH="GOARCH_${ARCH}"
```
jahkeup: I don't see the need for these to be passed by users as `--build-arg`, can you switch these to use `ENV` instead of `ARG`?

Dockerfile
```diff
@@ -495,6 +495,64 @@ RUN \
...
+USER builder
+WORKDIR /home/builder
+COPY ./hashes/govmomi ./hashes
+RUN \
+  sdk-fetch hashes && \
+  tar xf govmomi-${GOVMOMIVER}.tar.gz && \
+  rm govmomi-${GOVMOMIVER}.tar.gz && \
+  mkdir -p ${GOPATH}/src/github.com/vmware && \
+  mv govmomi-${GOVMOMIVER} ${GOPATH}/src/github.com/vmware/govmomi && \
+  ln -s ${GOPATH}/src/github.com/vmware/govmomi
```
jahkeup: Can we drop this? I don't see the path `$PWD?govmomi` used anywhere. Is it used implicitly somewhere?

jahkeup: I think this can be shrunk down a bit as there's little `cd`-ing:

1. create directory by setting it as the `WORKDIR`
2. download the tarball into that directory (within `$GOPATH`)
3. unpack the tarball with first path component stripped off (ie: dropping the `govmomi-X.Y.Z/` leading path component)

```suggestion
WORKDIR /home/builder/go/src/github.com/vmware/govmomi
COPY ./hashes/govmomi ./hashes
RUN \
  sdk-fetch hashes && \
  tar --strip-components=1 -xf govmomi-${GOVMOMIVER}.tar.gz && \
  rm govmomi-${GOVMOMIVER}.tar.gz
```

Dockerfile
```diff
@@ -495,6 +495,64 @@ RUN \
...
+ARG GOARCH_aarch64="arm64"
+ARG GOARCH_x86_64="amd64"
+ARG GOARCH_ARCH="GOARCH_${ARCH}"
+
+WORKDIR ${GOPATH}/src/github.com/vmware/govmomi
+COPY --chown=0:0 --from=sdk-license-scan /usr/libexec/tools/ /usr/libexec/tools/
+COPY --chown=0:0 --from=sdk-license-scan /home/builder/license-scan/clarify.toml .
+RUN \
+  cp -p LICENSE.txt /usr/share/licenses/govmomi && \
+  go mod tidy && \
```
jahkeup: Did you find `go mod tidy` to be required? I'm hesitant to do this in practice as it could modify the dependencies declared by upstream (and pull in more or less than what was originally expected to be).

Dockerfile
```diff
@@ -495,6 +495,64 @@ RUN \
...
+RUN \
+  mkdir -p /usr/share/licenses/govmomi && \
+  chown -R builder:builder /usr/share/licenses/govmomi
+
+ARG ARCH
+ARG TARGET="${ARCH}-bottlerocket-linux-gnu"
+ARG GOVMOMIVER="0.26.0"
+ARG GOVMOMISHORTCOMMIT="34586b6"
+ARG GOROOT="/home/builder/go"
+ARG GOPATH="${GOROOT}/go"
```
jahkeup: I'm not a fan of pushing non-stdlib sources into `GOROOT` this way. This implicitly supplies the package globally across all compilations which use the same `GOROOT`, rather than the usual `GOPATH` namespacing.

Can we instead use something like:

 ```suggestion
ARG GOPATH="${GOROOT}/gopath"
```

File: Dockerfile [hunk 5]
```diff
@@ -495,6 +495,64 @@ RUN \
...
+WORKDIR ${GOPATH}/src/github.com/vmware/govmomi/govc
+RUN \
+  export GOARCH="${!GOARCH_ARCH}" ; \
+  export PATH="${GOROOT}/bin:${PATH}" ; \
+  export GOFLAGS="-mod=vendor" ; \
+  go build -ldflags " \
+    -X ${PKGPATH}.BuildVersion=${GOVMOMIVER} \
+    -X ${PKGPATH}.BuildCommit=${GOVMOMISHORTCOMMIT} \
+    -X ${PKGPATH}.BuildDate=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
+    "
```
jahkeup: I like the symmetry of the paths elsewhere where the src & dst file paths are the same - this makes the provenance a bit more clear when tracing back through (more explicit!). What do you think about explicitly building the executable into the final path for later `COPY` use:

```suggestion
  mkdir -p /usr/libexec/tools
  go build -mod=vendor -o /usr/libexec/tools/govc -ldflags " \
    -X ${PKGPATH}.BuildVersion=${GOVMOMIVER} \
    -X ${PKGPATH}.BuildCommit=${GOVMOMISHORTCOMMIT} \
    -X ${PKGPATH}.BuildDate=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
    "
```

Dockerfile
```diff
@@ -495,6 +495,64 @@ RUN \
...
+  /usr/libexec/tools/bottlerocket-license-scan \
+    --clarify clarify.toml \
+    --spdx-data /usr/libexec/tools/spdx-data \
+    --out-dir /usr/share/licenses/govmomi \
+    go-vendor vendor
+
+WORKDIR ${GOPATH}/src/github.com/vmware/govmomi/govc
+RUN \
+  export GOARCH="${!GOARCH_ARCH}" ; \
+  export PATH="${GOROOT}/bin:${PATH}" ; \
```
jahkeup: Are there installed tools that need to be executed from `PATH` that exist only in `$GOROOT/bin`? If not, can we drop this line?

Dockerfile
```diff
@@ -495,6 +495,64 @@ RUN \
...
+  go mod vendor && \
+  /usr/libexec/tools/bottlerocket-license-scan \
+    --clarify clarify.toml \
+    --spdx-data /usr/libexec/tools/spdx-data \
+    --out-dir /usr/share/licenses/govmomi \
+    go-vendor vendor
+
+WORKDIR ${GOPATH}/src/github.com/vmware/govmomi/govc
+RUN \
+  export GOARCH="${!GOARCH_ARCH}" ; \
```
jahkeup: Nice shell indirect ;)

Dockerfile
```diff
@@ -495,6 +495,64 @@ RUN \
...
+  sdk-fetch hashes && \
+  tar xf govmomi-${GOVMOMIVER}.tar.gz && \
+  rm govmomi-${GOVMOMIVER}.tar.gz && \
+  mkdir -p ${GOPATH}/src/github.com/vmware && \
+  mv govmomi-${GOVMOMIVER} ${GOPATH}/src/github.com/vmware/govmomi && \
+  ln -s ${GOPATH}/src/github.com/vmware/govmomi
+
+ARG GOOS="linux"
+ARG CGO_ENABLED=0
+ARG PKGPATH="github.com/vmware/govmomi/govc/flags"
```
jahkeup: nit: this name confused me, can we call this after the aim? And maybe move closer to that section of the build rather than early on, here?

```suggestion
ARG BUILD_VERSION_PKG="github.com/vmware/govmomi/govc/flags"
```

jpculp: I was trying to preserve more or less what's inside the upstream [.goreleaser.yml](https://github.com/vmware/govmomi/blob/master/.goreleaser.yml), which is the package's recommended method of building from source.

Dockerfile
jpculp: I was trying to stay consistent with the `ARG`s in `sdk-go`, but I don't have a strong opinion on it.

Dockerfile
jpculp: The symlink? I only reference it towards the bottom when it's copying files since I'm hoping at some point in the future govmomi can just be built from `/home/builder/govmomi`.

Dockerfile
jpculp: No, the `go mod tidy` can safely be removed.

Dockerfile
jpculp: I wasn't crazy about it either, but (unless I misread) that is what we have going on in `sdk-go`. I'd like to stay consistent, so I think we should change it in `sdk-go` as well.

Dockerfile
jpculp: By building directly to `/usr/libexec/tools/govc` like in your other suggestion, this symlink isn't necessary anymore.

Dockerfile
jahkeup: I think the only one should consider exposing to users is `GOOS` which would allow building the toolchain on other platforms - but even that isn't terribly helpful as the SDK isn't set up to build for anything other than the equivalent to `GOOS=linux` at this time.

Having the likes of `GOARCH_x86_64` exposed to users permits misconfiguration and confusion:

```bash
# Build on x86_64 for RISC-V 64
[[ "$(uname -om)" = "x86_64 GNU/Linux" ]] &&
  docker build -t my-sdk:build --build-arg GOOS=freebsd --build-arg GOARCH_x86_64=riscv64 .
```

I'm not sure we'll have a case where overriding _some_ of these will make a future change easier or harder. I suggest that we hide `CGO_ENABLED`, `PKGPATH`, and `GOARCH_*`. Maybe also `GOOS`, though I'm not convinced either way yet.

Dockerfile
jahkeup: The `sdk-go` case is writing out the Go installation, right? Or are you seeing something else? In the case of the installation root, that's where Go's stdlib sources & intermediates should be installed.

Dockerfile
jahkeup: Ah, I wouldn't worry too much about matching the names. The same tactic is taken here as used by `goreleaser` and the method supports using a more context appropriate (to us & our usage) name.

My opinion is to choose our own name, but I'll leave it up to you, its a nit :)

jpculp:
Addressed @jahkeup's feedback.

jpculp:
Variable name tweak and fixed push. Proper diff [here](https://github.com/bottlerocket-os/bottlerocket-sdk/compare/ec995611b3ab0747decf7fc6689552d5ae4e4e3b..b63e06bb0815a7f30d3418c1b6ee3ff2766dfcb3).

jpculp:
Set an absolute path for the `hashes` file to stay consistent and avoid potential collisions.

jpculp:
- Removed GOPROXY="off" from `sdk-go`.
- Removed GOSUMB="off" from `sdk-go`.
- Change extraction point of go sdk to `/home/builder/sdk-go`.
- Changed GOROOT in `sdk-go` to `/home/builder/sdk-go`.
- Changed GOPATH in `sdk-go` to `/home/builder/gopath`.
- Set GOOS, GOROOT, GOPATH, and PATH in ENV in `sdk-go` to share with `sdk-govc`.
- Removed the now unnecessary bits from `sdk-govc`.

Dockerfile
```diff
@@ -495,6 +493,54 @@ RUN \
...
+COPY --chown=0:0 --from=sdk-license-scan /usr/libexec/tools/ /usr/libexec/tools/
+COPY --chown=0:0 --from=sdk-license-scan /home/builder/license-scan/clarify.toml .
+RUN \
+  cp -p LICENSE.txt /usr/share/licenses/govmomi && \
+  go mod vendor && \
+  /usr/libexec/tools/bottlerocket-license-scan \
+    --clarify clarify.toml \
+    --spdx-data /usr/libexec/tools/spdx-data \
+    --out-dir /usr/share/licenses/govmomi \
+    go-vendor vendor
```
jahkeup: nit: my eyes struggled to find the directory from the sub-command, maybe make this clearly a relative directory usage like this?

```suggestion
    go-vendor ./vendor
```

Dockerfile
```diff
@@ -495,6 +493,54 @@ RUN \
...
+WORKDIR ${GOPATH}/src/github.com/vmware/govmomi/govc
+RUN \
+  export GOARCH="${!GOARCH_ARCH}" ; \
+  export CGO_ENABLED=0 ; \
+  export BUILD_VERSION_PKG="github.com/vmware/govmomi/govc/flags" ; \
+  go build -mod=vendor -o /usr/libexec/tools/govc -ldflags " \
+    -X ${BUILD_VERSION_PKG}.BuildVersion=${GOVMOMIVER} \
+    -X ${BUILD_VERSION_PKG}.BuildCommit=${GOVMOMISHORTCOMMIT} \
+    -X ${BUILD_VERSION_PKG}.BuildDate=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
+    "
```
jahkeup: nit: I'd prefer to see the targeted package to be compiled. Especially since we're reconstituting a `GOPATH` this could be the full pkg path to the final bin's `main`:

```suggestion
  go build -mod=vendor -o /usr/libexec/tools/govc -ldflags " \
    -X ${BUILD_VERSION_PKG}.BuildVersion=${GOVMOMIVER} \
    -X ${BUILD_VERSION_PKG}.BuildCommit=${GOVMOMISHORTCOMMIT} \
    -X ${BUILD_VERSION_PKG}.BuildDate=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
    " github.com/vmware/govmomi/govc
```

Dockerfile
```diff
@@ -550,12 +596,12 @@ COPY --chown=0:0 --from=sdk-rust \
   /usr/share/licenses/rust/
 
 # "sdk-go" has the Go toolchain and standard library builds.
-COPY --chown=0:0 --from=sdk-go /home/builder/go/bin /usr/libexec/go/bin/
-COPY --chown=0:0 --from=sdk-go /home/builder/go/lib /usr/libexec/go/lib/
-COPY --chown=0:0 --from=sdk-go /home/builder/go/pkg /usr/libexec/go/pkg/
-COPY --chown=0:0 --from=sdk-go /home/builder/go/src /usr/libexec/go/src/
+COPY --chown=0:0 --from=sdk-go /home/builder/sdk-go/bin /usr/libexec/go/bin/
```
jahkeup: Nice! I like the separate namespacing of `sdk-go` for the SDK's Go toolchain and especially so because it drives away from the collision of default `GOPATH` (which would be `/home/builder/go` and even explicitly used this way before).

Dockerfile
```diff
@@ -402,27 +402,25 @@ ARG CGO_CFLAGS="${CFLAGS}"
...
 
 # Build the standard library with and without PIE. Target binaries
 # should use PIE, but any host binaries generated during the build
 # might not.
-WORKDIR /home/builder/go
+WORKDIR /home/builder/sdk-go
+ENV GOOS="${GOOS}" \
+  GOROOT="/home/builder/sdk-go" \
+  GOPATH="/home/builder/gopath" \
+  PATH="/home/builder/sdk-go/bin:${PATH}"
```
jahkeup: Maybe a quick comment to explain why these are set this way and here. Also, most importantly, why consuming stages probably _shouldn't_ need to set any of them.

```suggestion
# SDK Go toolchain environment; inherited by later build stages
ENV GOOS="${GOOS}"
ENV GOROOT="/home/builder/sdk-go"
ENV GOPATH="/home/builder/gopath"
# PATH is prefixed to select it over the host's Go tools.
ENV PATH="/home/builder/sdk-go/bin:${PATH}"
```

Dockerfile
jahkeup: I'm okay with this change, but I'm less certain this should accompany the `govc` addition. 

If no one else has objections, then it can come in this way, but I'd prefer to have seen this in its own PR (even if we'd have to review & merge that before `govc`).

jahkeup (approved):
All goods things here 👍🏽 I have a couple suggestions and questions for the gallery to weigh in on (regarding updates to clean up `sdk-go`).

I'll be AFK for a bit - so please consider this my continued approval barring any major updates :)

Dockerfile
jpculp: That is certainly clearer. Since our documentation just has `go-vendor vendor`, I'm a bit torn.

Dockerfile
jpculp: If we go with this, we probably don't have to change workdirs either.

webern (approved):

Dockerfile
```diff
@@ -402,27 +402,25 @@ ARG CGO_CFLAGS="${CFLAGS}"
...
 RUN \
-  export GOARCH="${!GOARCH_ARCH}" ; \
```
bcressey: This doesn't look correct to me. IIRC it means the installed standard library will be built for the host and not the target arch, and would need to be rebuilt by each Go build in Bottlerocket proper.

Dockerfile
```diff
@@ -495,6 +493,54 @@ RUN \
...
+  go mod vendor && \
+  /usr/libexec/tools/bottlerocket-license-scan \
+    --clarify clarify.toml \
+    --spdx-data /usr/libexec/tools/spdx-data \
+    --out-dir /usr/share/licenses/govmomi \
+    go-vendor vendor
+
+WORKDIR ${GOPATH}/src/github.com/vmware/govmomi/govc
+RUN \
+  export GOARCH="${!GOARCH_ARCH}" ; \
```
bcressey: This seems backwards; we'll end up building `govc` for the target architecture rather than the host, but we need to run it as a host program.

Dockerfile
```diff
@@ -495,6 +493,54 @@ RUN \
...
+  sdk-fetch /home/builder/hashes && \
+  tar --strip-components=1 -xf govmomi-${GOVMOMIVER}.tar.gz && \
+  rm govmomi-${GOVMOMIVER}.tar.gz
+
+ARG GOARCH_aarch64="arm64"
+ARG GOARCH_x86_64="amd64"
+ARG GOARCH_ARCH="GOARCH_${ARCH}"
+
+COPY --chown=0:0 --from=sdk-license-scan /usr/libexec/tools/ /usr/libexec/tools/
+COPY --chown=0:0 --from=sdk-license-scan /home/builder/license-scan/clarify.toml .
```
bcressey: Not sure how this could be the correct `clarify.toml` - it's the one for `license-scan`, which is a Rust program with a very different set of dependencies. Possibly it's not used?

Dockerfile
```diff
@@ -402,27 +402,25 @@ ARG CGO_CFLAGS="${CFLAGS}"
...
 RUN ./make.bash --no-clean
 
 # Build the standard library with and without PIE. Target binaries
 # should use PIE, but any host binaries generated during the build
 # might not.
-WORKDIR /home/builder/go
+WORKDIR /home/builder/sdk-go
+ENV GOOS="${GOOS}" \
+  GOROOT="/home/builder/sdk-go" \
+  GOPATH="/home/builder/gopath" \
```
bcressey: Does building `go` with these values set have any consequences for the final binary - do they override what would be the default values?

bcressey (changes requested):

Dockerfile
jpculp: I don't believe this has final binary implications, but without them subsequent `go` builds have a hard time using our build of `go`. If `GOROOT` is not specified then it will use the `go` that was installed by `dnf`. As for `GOPATH`, the default is `$HOME/go` which is at odds with our current `GOROOT` location.

Dockerfile
jpculp: From what I can tell, it is checking and outputting licenses for dependencies correctly to `/usr/share/licenses/govmomi/`.

Dockerfile
jpculp: In that case we can just remove all the ARCH stuff from the `sdk-govc` block, right?

jpculp:
- Re-added `export GOARCH="${!GOARCH_ARCH}"` to `sdk-go` to fix cross compiles.
- Removed extraneous ARCH variables from `sdk-govc` as `govc` binary needs to match the container arch, not the target arch.
- Specified path to `govc` for `go-build` so that we don't need to change WORKDIRs so much.
- Switched this PR to a draft so that the `sdk-go` changes can be moved to their own PR (https://github.com/bottlerocket-os/bottlerocket-sdk/pull/55).

Dockerfile
jahkeup: #55 has the sdk-go changes separately now - will move discussion over there.

jpculp:
Rebased post-#55 merger.

Dockerfile
jahkeup: 🤷🏽 the usage can be different here and revised in the docs (if that's appropriate) to make it clearer that a directory is being passed as an argument. I'd vote to add the `./`, it is ultimately up to you here. It's only a nit :)

Dockerfile
```diff
@@ -494,6 +494,46 @@ RUN \
...
+    --out-dir /usr/share/licenses/govmomi \
+    go-vendor vendor
+
+RUN \
+  export CGO_ENABLED=0 ; \
+  export BUILD_VERSION_PKG="github.com/vmware/govmomi/govc/flags" ; \
+  go build -mod=vendor -o /usr/libexec/tools/govc -ldflags " \
+    -X ${BUILD_VERSION_PKG}.BuildVersion=${GOVMOMIVER} \
+    -X ${BUILD_VERSION_PKG}.BuildCommit=${GOVMOMISHORTCOMMIT} \
+    -X ${BUILD_VERSION_PKG}.BuildDate=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
```
jahkeup: It's worth noting that using a moving date (time of build when not cached) will make creating a reproducible-build of this executable a challenge. Can the date instead be made into a static argument alongside the `GOVMOMIVER` and `GOVMOMISHORTCOMMIT` like `GOVMOMIDATE`? The value given could either be the release's timestamp or the timestamp that the sources are updated by a human.

jahkeup (approved):

zmrow (approved):
:waffle:

Dockerfile
jpculp: Others might disagree, but to me the build date should be the date that the binary is actually built. It's absolutely not reproducible, but I think that's kind of the point. It's signifying that it isn't the exact binary that was built last week, or the week before (although hopefully it is functionally the same!). I don't think this should confuse anyone since it's not even visible with `govc version`. You can only see the date when you run `govc version -l`.

jpculp:
All 4 builds tested. Also built bottlerocket on both architectures. Ready for review!

Dockerfile
jahkeup: I see the build date giving users significantly less clarity on the artifact's origin in comparison to a timestamp related to a build's source revision. And, to me, the build date - specifically without a source-related date - seems like a misleading reference for users. It is very easy to mistake a recent date in version data meaning that its the product of a more recent release being consumed. That is, if the source revision is unchanged but the build date is, then users can easily (when looking for it) find a recent looking date but actually with a potentially much older version of the tool. I'd much rather see the "capture date" of when we updated the recipe or the source revision's original timestamp used over the arbitrary build timestamp.

If the date string isn't easily discoverable (and unlikely to be accessed) then perhaps we ought to omit it rather than offer a potentially confusing timestamp.

(in my view, this point of contention is analogous to `#!/bin/bash` v. `#!/usr/bin/env bash` - there are subtle differences in some contexts and profound in [others, that don't have a `/bin/bash`](https://nixos.org)!)

Dockerfile
bcressey: If you omit the clarify argument, does it still work?

Dockerfile
jpculp: After some further discussion, let's go with the upstream commit date.

jpculp:
- Removed `clarify.toml` as it was unnecessary.
- Changed `govc` BuildDate to a hard coded value that for now will be tied to the upstream commit date.
- Added `./` to make it more apparent that the vendor in `go-vendor` refers to a directory.

bcressey (approved):

Issue #61, opened 2021-11-01
update to Rust 1.56.1+


bcressey:
[Rust 1.56.1](https://blog.rust-lang.org/2021/11/01/Rust-1.56.1.html) is out with a mitigation for CVE-2021-42574.

PR #62, opened 2021-11-10
minor update to SDK dependencies

jpculp:
**Issue number:**

#61 

**Description of changes:**

This is a minor point release **(v0.23.1)** containing mostly security fixes.

* Rust: **1.56.0 → 1.56.1**
* Go: **1.16.9 → 1.16.10**
* govc: **0.27.0 → 0.27.1**

**Testing done:**

- [x] Build SDK (and Bottlerocket) `x86_64` on `x86_64`
- [x] Build SDK (and Bottlerocket) `aarch64` on `x86_64`
- [x] Build SDK (and Bottlerocket) `aarch64` on `aarch64`
- [x] Build SDK (and Bottlerocket) `x86_64` on `aarch64`

**Terms of contribution:**

### DIFF
--- a/hashes/rust
+++ b/hashes/rust
@@ -1,6 +1,6 @@
-# https://static.rust-lang.org/dist/rustc-1.56.0-src.tar.xz
-SHA512 (rustc-1.56.0-src.tar.xz) = 2daa365524b47dcc48e49a0e9c8c45988af44c0845e2695dc5053f18e768e49acf3dbdd77f808dbf260546ef608eb47c593544012dd05675cb7e6b6223900315
-### See https://github.com/rust-lang/rust/blob/1.56.0/src/stage0.txt for what to use below. ###
+# https://static.rust-lang.org/dist/rustc-1.56.1-src.tar.xz
+SHA512 (rustc-1.56.1-src.tar.xz) = 193468e211cde9ebc5f6e30b8e3733b79bd8710fe6dd45c7ed8d4392f91010d30466787afd4d0b2041cd7dd994924fee8ad111048824e248bd994959e55bf15f
+### See https://github.com/rust-lang/rust/blob/1.56.1/src/stage0.txt for what to use below. ###
 # https://static.rust-lang.org/dist/2021-09-09/rust-std-1.55.0-x86_64-unknown-linux-gnu.tar.xz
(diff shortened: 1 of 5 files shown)


bcressey (approved):

hashes/rust
```diff
@@ -1,6 +1,6 @@
-# https://static.rust-lang.org/dist/rustc-1.56.0-src.tar.xz
-SHA512 (rustc-1.56.0-src.tar.xz) = 2daa365524b47dcc48e49a0e9c8c45988af44c0845e2695dc5053f18e768e49acf3dbdd77f808dbf260546ef608eb47c593544012dd05675cb7e6b6223900315
-### See https://github.com/rust-lang/rust/blob/1.56.0/src/stage0.txt for what to use below. ###
+# https://static.rust-lang.org/dist/rustc-1.56.1-src.tar.xz
+SHA512 (rustc-1.56.1-src.tar.xz) = 193468e211cde9ebc5f6e30b8e3733b79bd8710fe6dd45c7ed8d4392f91010d30466787afd4d0b2041cd7dd994924fee8ad111048824e248bd994959e55bf15f
+### See https://github.com/rust-lang/rust/blob/1.56.1/src/stage0.txt for what to use below. ###
```
tjkirch, line 3: This is worrying me a bit.  Why wouldn't any of tarballs below change?

Looks like stage0.txt didn't change at all between the 1.56.0 and 1.56.1 trees.  They've gotten rid of stage0.txt in the live tree, switching to stage0.json instead.  I wonder if they're not updating stage0.txt anymore.

I think we should find some way to confirm this is correct, given that (I think) the build process uses the tarballs below, and they didn't change..?

webern (approved):

hashes/rust
jpculp: The difference between 1.56.0 and 1.56.1 is pretty much a single commit https://github.com/rust-lang/rust/commit/dd61274930ec0cd17711fab52d2bc9ad3e9053de so I wasn't terribly surprised to see the same stage0.txt as the 1.56.0 release.

tjkirch: Oh, it's because the tarballs below are the 1.55 tarballs, used as stage0 to build 1.56, and 1.55 hasn't changed.  Just this source tarball of 1.56 would change for a point release.  Same thing happened with the 1.45.x point releases.

tjkirch (approved):

cbgbt (approved):

Issue #57, opened 2021-06-30
Document multi-arch build and publish steps


webern:
The steps to build the SDK for both host architectures and both target architectures (four builds in all) are complicated. Additionally for use by Bottlerocket we also combine the host architectures into a multi-architecture tag/manifest.

The steps and scripts that we use to do this internally should be more-or-less identical to what anyone else would do if they wanted to build and publish the SDK to their own container image repos. Let's document these steps and include the helper script here.

jpculp:
Opened PR for documentation. Will work on potential helper script in a later PR.

PR #63, opened 2021-11-10
Add publish target to Makefile

jpculp:
**Issue number:**

#57 

**Description of changes:**

This adds a helper script `publish-sdk` to publish the SDK and toolchain images to ECR and generate manifests if necessary. It can be called on its own or via the Makefile.

To publish via the Makefile, you will need to run:
```
make publish \
    REGISTRY=aws_account_id.dkr.ecr.us-east-1.amazonaws.com \
    SDK_NAME=my-custom-bottlerocket
```

**Testing done:**

Pushed a local SDK build to ECR.

**Terms of contribution:**


webern (approved):
Nice.

jpculp:
Added `print-version` to **.PHONY**

sdk2ecr
```diff
@@ -0,0 +1,114 @@
...
+   case "${opt}" in
+      --version=*) VERSION="${optarg}" ;;
+      --registry=*) REGISTRY="${optarg}" ;;
+      --sdk-name=*) SDK_NAME="${optarg}" ;;
+   esac
+done
+
+if [ -z "${VERSION}" ]; then
+  VERSION=$(make print-version | tail -1)
+fi
```
bcressey: I'd prefer to just treat `VERSION` as required too - it's odd to have a tool called by `make` that calls back into `make`.

sdk2ecr
```diff
@@ -0,0 +1,114 @@
...
+   optarg="$(expr "${opt}" : '[^=]*=\(.*\)')"
+   case "${opt}" in
+      --version=*) VERSION="${optarg}" ;;
+      --registry=*) REGISTRY="${optarg}" ;;
+      --sdk-name=*) SDK_NAME="${optarg}" ;;
+   esac
+done
+
+if [ -z "${VERSION}" ]; then
+  VERSION=$(make print-version | tail -1)
```
jpculp: Alternatively, we can define the version in a **VERSION** file, rather than the Makefile, and remove the `print-version` target.

sdk2ecr
```diff
@@ -0,0 +1,114 @@
...
+    return
+  fi
+
+  local image="${remote_image#*/}"
+  local version="${image##*:}"
+  local repo="${image%:*}"
+
+  aws ecr batch-delete-image \
+    --repository-name "${repo}" \
+    --image-id imageTag="${version}"
```
bcressey: I have this change in my local copy of the script; I hit some weird errors otherwise even when the image existed and should have been deleted:

```suggestion
  aws ecr batch-delete-image \
    --repository-name "${repo}" \
    --image-id imageTag="${version}" \
    --region us-east-1
```

We might want to suppress this output since the error messages are a little confusing.

sdk2ecr
```diff
@@ -0,0 +1,114 @@
...
+   esac
+done
+
+if [ -z "${VERSION}" ]; then
+  VERSION=$(make print-version | tail -1)
+fi
+
+for required_arg in VERSION REGISTRY SDK_NAME; do
+  [ -z "${!required_arg}" ] && echo "${required_arg} not set!" >&2 && exit 1
+done
```
bcressey: It'd also be good to add defaults for these to the `Makefile` even though we also check here. Maybe I'm weird but I would tend to read the `Makefile` ahead of time to figure out what I should specify in my `make` invocation, and wouldn't think to look into the various scripts until I hit an error.

We could default to `bottlerocket` as the SDK name and `localhost:5000` for the registry.

jpculp:
- Move VERSION variable out of the Makefile and into a VERSION file.
- Add required argument tests in lieu of default values in `publish` target.
- Change AWS CLI command based on whether REGISTRY is ECR or ECR Public.
- Fix bash substitution that wasn't properly parsing the repo from ECR Public registries.

jpculp:
- Remove the block to fetch version from VERSION file if `--version` is not passed.
- Switch up ECR logic to guarantee the `aws ecr public` command runs with the `us-east-1` region.
- Create the "${HOME}/.docker/manifests" directory if it does not exist to avoid an early exit.
- Bump VERSION file.

jpculp:
- Renamed `sdk2ecr` to `publish-sdk`).
- Removed unneeded `/` after `$(TOP)`.
- Rebased to the latest develop.

sdk2ecr
jpculp: Rather than suppressing, the errors were mitigated.

webern (approved):

publish-sdk
```diff
@@ -0,0 +1,119 @@
...
+  local repo="${image%:*}"
+
+  # clean up any remote (ECR) copy of the manifest
+  if [[ "${REGISTRY}" =~ ^public\.ecr\.aws\/ ]]; then
+    aws ecr-public batch-delete-image \
+      --repository-name "${repo}" \
+      --image-ids imageTag="${VERSION}" \
+      --region "us-east-1"
+  elif [[ "${REGISTRY}" =~ ^[0-9]+\.dkr\.ecr\.[a-z0-9-]+\.amazonaws\.com$ ]]; then
+    aws ecr-public batch-delete-image \
```
bcressey: Shouldn't this be ECR private?
```suggestion
    aws ecr batch-delete-image \
```

jpculp: Fixed and re-tested with ECR private!

bcressey (approved):

webern (approved):

PR #66, opened 2021-12-14
Add license-tool

arnaldo2792:
**Issue number:**
N / A

**Description of changes:**

```
ec21709 Add license-tool
```

There are cases where multiple licenses could apply to a package in the  main bottlerocket repo, depending on who will distribute the software. For such packages, the licensing information has to be provided at builtime, but can't be hardcoded in the RPM recipies.

 `license-tool` can be used to retrieve the license information for a package with multiple distribution licenses. It uses a toml configuration file with the licenses that apply to a set of packages. This allows Bottlerocket's users to build the OS packages with the license that better applies to them.

**Testing done:**
There are unit tests that cover how the tool works

**Terms of contribution:**


license-tool/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+    Ok(())
+}
+
+/// Fetches the licenses in the `PackageLicense` object, and creates a copy of them in `dest`
+async fn fetch_licenses<P>(package_license: &PackageLicense, dest: P) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    for license in &package_license.licenses {
+        let path: PathBuf = get_path(&license, &dest)?.into();
```
jpculp:
```suggestion
        let path: PathBuf = get_path(license, &dest)?.into();
```
clippy says this is a needless borrow (same with lines 188, 298, and 309).

license-tool/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+fn replace_env_vars<S>(path: S) -> Result<Url>
+where
+    S: AsRef<str>,
+{
+    let mut segments: Vec<String> = vec!["file://".to_string()];
+    let path = path.as_ref().to_string();
+    let regex = Regex::new(ENV_PATTERN)?;
+    let file_path = path.split("://").skip(1).next().expect("Invalid path");
+
+    for segment in file_path.split("/") {
```
jpculp:
```suggestion
    for segment in file_path.split('/') {
```
This can be a Char.

license-tool/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+
+/// Replaces the ENV variables when present in a file URL
+fn replace_env_vars<S>(path: S) -> Result<Url>
+where
+    S: AsRef<str>,
+{
+    let mut segments: Vec<String> = vec!["file://".to_string()];
+    let path = path.as_ref().to_string();
+    let regex = Regex::new(ENV_PATTERN)?;
+    let file_path = path.split("://").skip(1).next().expect("Invalid path");
```
jpculp:
```suggestion
    let file_path = path.split("://").nth(1).expect("Invalid path");
```
Another clippy suggestion.

license-tool/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+/// Prints the spdx id for the package
+fn print_spdx_id<S>(packages_licenses: &HashMap<String, PackageLicense>, package: S) -> Result<()>
+where
+    S: AsRef<str>,
+{
+    let package = package.as_ref();
+    let package_license = packages_licenses.get(package).context(format!(
+        "Couldn't find configuration for package '{}'",
+        package
+    ))?;
```
jpculp, line 95: I would recommend taking advantage of snafu. You can share the error message with L112.

license-tool/README.md
```diff
@@ -0,0 +1,39 @@
+# Bottlerocket's license-tool helper
+There are cases where multiple licenses could apply to a package in the main bottlerocket repo, depending on who will distribute the software (a final user vs a corporate).
+For such packages, the users should have the ability to chose what license they want to use for a package with multiple distribution licenses, without hardcoding this information in the RPM recepies.
```
jpculp:
```suggestion
For such packages, the users should have the ability to choose what license they want to use for a package with multiple distribution licenses, without hardcoding this information in the RPM recipes.
```

license-tool/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+The **license-tool** helper can be used to select the licenses that will apply to a set of packages.
+The tool provides commands to retrieve the licenses' spdxid and files paths, which are used in the RPM spec recipies.
```
jpculp:
```suggestion
The tool provides commands to retrieve the licenses' SPDX id and file paths, which are used in the RPM spec recipes.
```

license-tool/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+The Bottlerocket build system uses this helper to let users select the license that best applies to them, for some of the third-party packages distributed in the base image.
+
+## Use in the Bottlerocket build system
+The Bottlerocket build system calls `bottlerocket-license-tool fetch` to create a local copy of the licenses files, before the packages builds begin.
+In the spec file of a package with multiple distribution licenses, `bottlerocket-license-tool` is called with the `spdx-id` and `path` commands to fetch the license's SDPX id and file path.
+For example, a RPM spec recipie could have:
```
jpculp:
```suggestion
For example, a RPM spec recipe could have:
```

Dockerfile
```diff
@@ -532,6 +532,22 @@ RUN \
 
 # =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
+FROM sdk-rust as sdk-license-tool
+
+USER root
+COPY license-tool /home/builder/license-tool
+RUN \
+  mkdir -p /usr/libexec/tools /home/builder/license-tool /usr/share/licenses/bottlerocket-license-tool && \
+  chown -R builder:builder /usr/libexec/tools /home/builder/license-tool /usr/share/licenses/bottlerocket-license-tool
```
jpculp: It seems a little odd to COPY to a location and then immediately `mkdir -p` the same location. I suspect the COPY needs to come after `WORKDIR /home/builder/license-tool`

Dockerfile
```diff
@@ -532,6 +532,22 @@ RUN \
...
+
+USER builder
+WORKDIR /home/builder/license-tool
+RUN cargo build --release --locked
+RUN install -p -m 0755 target/release/bottlerocket-license-tool /usr/libexec/tools/
+COPY COPYRIGHT LICENSE-APACHE LICENSE-MIT /usr/share/licenses/bottlerocket-license-tool/
```
bcressey: This should look more like the `cargo-deny` build, where we have a `clarify.toml` and scan the project with our own license tool in order to generate the attribution file.

license-tool/README.md
```diff
@@ -0,0 +1,39 @@
...
+# ...
+
+%files my-package
+%license %{license_file}
+```
+
+## Licenses file example
+
+```toml
+[my-pacakge]
```
bcressey:
```suggestion
[my-package]
```

license-tool/README.md
```diff
@@ -0,0 +1,39 @@
...
+spdx-id = "SPDX-ID AND SPDX-ID-2" # Package with multiple licenses
+licenses = [
+  { license-url = "file:///path/to/spdx-id-license.txt" }, # This file is copied from a file system
+  { license-url = "https://www.cool-licenses.com/spdx-id-license-v2.txt", path = "spdx-id-license-2.txt" } # This file is fetch from an https endpoint
```
bcressey: Maybe use an example domain so that we don't end up linking to a bad site:
```suggestion
  { license-url = "https://localhost/spdx-id-license-v2.txt", path = "spdx-id-license-2.txt" } # This file is fetch from an https endpoint
```

bcressey: Will it do the right thing if the path is not specified - use the remote file name as determined from the URL?

arnaldo2792:
Forced push includes:

- Fixed typos
- Fixed clippy warnings
- Move `COPY` directive in Dockerfile

license-tool/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+    #[serde()]
+    licenses: Vec<License>,
+}
+
+/// Holds the configurations for a license
+#[derive(Deserialize, Debug, Clone)]
+struct License {
+    #[serde(rename(deserialize = "license-url"))]
+    license_url: Url,
+    #[serde()]
```
bcressey: What's this for?

license-tool/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+) -> Result<()>
+where
+    S: AsRef<str>,
+    P: AsRef<Path>,
+{
+    let package_name = package_name.as_ref();
+    let package_license = packages_licenses.get(package_name).context(format!(
+        "Couldn't find configuration for package '{}'",
+        package_name
+    ))?;
```
bcressey, line 113: Can you confirm that anyhow context errors go to stderr?

license-tool/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+    let licenses_file = licenses_file.as_ref();
+    Ok(toml::from_slice(&fs::read(&licenses_file).context(
+        format!("Failed to read file '{}'", licenses_file.display()),
+    )?)?)
+}
+
+/// Replaces the ENV variables when present in a file URL
+fn replace_env_vars<S>(path: S) -> Result<Url>
+where
+    S: AsRef<str>,
```
bcressey: Why do we need this functionality?

I'd probably look at using the `envsubst` library instead if we really have to do this.

license-tool/README.md
arnaldo2792: Yes, I'll add some comments to make that clear. If `path` is omitted, the file name will be determined from the URL.

license-tool/src/main.rs
arnaldo2792: Yes:

```shell
[I]  > cargo run -- -l Not-existing.toml fetch -d ./blah
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/bottlerocket-license-tool -l Not-existing.toml fetch -d ./blah`
# Error from context
Error: Failed to read file 'Not-existing.toml'

Caused by:
    No such file or directory (os error 2)
[I] > cargo run -- -l Not-existing.toml fetch -d ./blah 2>/dev/null
# Nothing printed
```

license-tool/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+    let mut all_paths = Vec::new();
+    for license in &package_license.licenses {
+        all_paths.push(get_path(license, &dest)?);
+    }
+
+    Ok(all_paths)
+}
+
+/// Returns the path to the destination file for the license
+fn get_path<P>(license: &License, base: P) -> Result<String>
```
cbgbt: nit: I feel like `get_license_destination` or `get_license_destination` might be a better name for this. Same deal for `get_paths`

license-tool/src/main.rs
arnaldo2792: Talked about this offline, we agreed that for binaries we will favor `anyhow` for the moment.

cbgbt: Snafu is exceptional for libraries where your consumers may want a high degree of granularity over matching on errors that occur, but I think `anyhow` is pretty well-regarded for binaries. Nobody will see the nice error types from Snafu, only the error messages.

license-tool/src/main.rs
arnaldo2792: A bad engineer that doesn't know how `serde` works, I'll remove it in my next push.

license-tool/src/main.rs
cbgbt: Midair collision!

arnaldo2792:
Forced push includes:

- Removed `replace_env_vars` function, this functionality is not longer required
- Added `license-scan` for `license-tool`
- Per offline discussion with @bcressey , `license-url` is now optional, and we will only attempt to fetch the file if present; `path` is now required, and is used to read/write the file from/to `--sources`/`--destination`

license-tool/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "bottlerocket-license-tool"
+version = "0.1.0"
+authors = ["Arnaldo Garcia Rincon <agarrcia@amazon.com>"]
+edition = "2018"
+publish = false
+
+[dependencies]
+anyhow = "1"
+argh = "0.1.3"
```
bcressey: Why the older version of `argh`?

license-tool/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+use std::io;
+use std::path::{Path, PathBuf};
+use url::Url;
+
+/// Stores arguments
+#[derive(FromArgs, PartialEq, Debug)]
+struct Args {
+    /// configuration file with the licenses to be used
+    #[argh(option, short = 'l')]
+    licenses_file: String,
```
bcressey: Can we default to just `Licenses.toml` where it's assumed to be in the current working directory?

license-tool/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+    Fetch(FetchArgs),
+}
+
+/// Returns the spdx-id for the package
+#[derive(FromArgs, Debug, PartialEq)]
+#[argh(subcommand, name = "spdx-id")]
+struct SpdxIdArgs {
+    /// the package name used to look up for the licenses
+    #[argh(option, short = 'p')]
+    package_name: String,
```
bcressey: Stylistic nit: why make this an option if it's a required argument?

license-tool/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+}
+
+/// Creates a copy of the licenses files in the dest directory
+#[derive(FromArgs, Debug, PartialEq)]
+#[argh(subcommand, name = "fetch")]
+struct FetchArgs {
+    /// the destination folder for the licenses
+    #[argh(option, short = 'd')]
+    destination: PathBuf,
```
bcressey: Same nit here - if we only take one option we could just make it a positional argument.

license-tool/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+/// Prints out a space-separated list of the paths to the licenses files
+#[derive(FromArgs, Debug, PartialEq)]
+#[argh(subcommand, name = "path")]
+struct PathArgs {
+    /// the package name used to look up for the licenses
+    #[argh(option, short = 'p')]
+    package_name: String,
+    /// the source folder where the licenses are
+    #[argh(option, short = 's')]
+    source: PathBuf,
```
bcressey: Here I might make the package name a positional argument, and have "source" as a real option.

If it's `None`, we just print the license path names as is. If it's set, then we prefix each path name with that value.

It might be worth calling it "prefix" and using `-p | --prefix` for this option. "source" implies some level of validation to me, but we're really just pasting two tokens together.

arnaldo2792:
Forced push includes:

- Updated `argh` version
- Change optional parameters to positional parameters, as suggested

cbgbt (approved):

bcressey (approved):

PR #82, opened 2022-09-14
update sources and add new tools

### LINKED ISSUE #64: system directories should be owned by root
**Image I'm using:**
v0.23.1

**Issue or Feature Request:**
Spotted this in https://github.com/bottlerocket-os/bottlerocket/pull/1799#discussion_r748001352 - the SDK has some directories owned by `builder` that should be owned by `root`.

```
[builder@b9e9b9e0f2db /]$ ls -latr / | grep builder
drwxr-xr-x.   3 builder builder 4096 Nov 16 22:58 x86_64-bottlerocket-linux-musl
drwxr-xr-x.  14 builder builder 4096 Nov 16 22:58 usr
drwxr-xr-x.   3 builder builder 4096 Nov 16 23:01 x86_64-bottlerocket-linux-gnu
```

### LINKED ISSUE #75: Update base image to F36 or greater
**Image I'm using:** 0.26.0

**Issue or Feature Request:** [Updating `dbus-broker` in bottlerocket](https://github.com/bottlerocket-os/bottlerocket/issues/2175) requires a newer `meson` package than is available in our [current F35 base](https://github.com/bottlerocket-os/bottlerocket-sdk/blob/c726347768074d39d499c8e5b235a174756c1438/Dockerfile#L1).

To fix this, we should move to a newer base image.

bcressey:
**Issue number:**
Fixes #75
Fixes #64

**Description of changes:**
The first set of changes includes the regular updates - to the Fedora base image, Buildroot, Rust, Go, etc.

I also updated the Rust dependencies, with related changes to `clarify.toml` and `deny.toml`. For good measure I made sure that all the Rust tools are scanned with both `bottlerocket-license-scan` and `cargo deny`. This required a considerable amount of iteration, so I added some additional layers to speed up successive builds.

The second set of changes includes new tools to support Secure Boot. The main one is [aws-kms-pkcs11](https://github.com/JackOfMostTrades/aws-kms-pkcs11), a PKCS#11 provider using AWS KMS as the backend, along with tools like `pesign` for signing. The other tools are helpful either for troubleshooting or for generating the EFI variable data in the required formats.

Finally, I added the `dir2fs` tool to better optimize filesystem sizes, and fixed the permissions on the sysroot directories, so packages will now fail to build if they attempt to modify the filesystem outside the build root.

**Testing done:**
Built many Bottlerocket images for both architectures on both aarch64 and x86_64 hosts.

I also signed all the relevant files for Secure Boot and confirmed that I could boot with Secure Boot enabled on both architectures under QEMU.

**Terms of contribution:**


jpculp (approved):

Dockerfile
```diff
@@ -599,17 +587,96 @@ RUN \
...
+  LICENSE-APACHE LICENSE-MIT \
+  /usr/share/licenses/cargo-deny/
+
+COPY --chown=1000:1000 \
+  COPYRIGHT LICENSE-APACHE LICENSE-MIT \
+  /usr/share/licenses/bottlerocket-license-tool/
+
+COPY --chown=1000:1000 \
+  COPYRIGHT LICENSE-APACHE LICENSE-MIT \
+  /usr/share/licenses/bottlerocket-license-scan/
```
arnaldo2792, line 637: Shouldn't these also be owned by UID `0`? Aren't directories under `/usr/share/licenses/*` considered system directories?

Dockerfile
```diff
@@ -42,8 +42,8 @@ RUN \
   git config --global user.name "Builder" && \
   git config --global user.email "builder@localhost"
 
-ARG BRVER="2021.02.3"
-ARG KVER="5.4.125"
+ARG BRVER="2022.05.2"
+ARG KVER="5.10.129"
```
arnaldo2792, line 47: Is this version of the kernel tightly coupled with the buildroot version?

patches/buildroot/0001-disable-shared-for-host-builds-of-gmp-isl-mpc-mpfr.patch
```diff
@@ -1,7 +1,7 @@
-From ce9da3957e01aa81af49210bb7fec3d8cb440690 Mon Sep 17 00:00:00 2001
+From dc72132b09daa401e9631c9af6463d93f035682d Mon Sep 17 00:00:00 2001
 From: Ben Cressey <bcressey@amazon.com>
 Date: Sun, 6 Aug 2017 23:15:02 +0000
-Subject: [PATCH 1/5] disable shared for host builds of gmp, isl, mpc, mpfr
+Subject: [PATCH] disable shared for host builds of gmp, isl, mpc, mpfr
```
arnaldo2792, line 4: Seems odd that `git` numbered the patch but not the subject :sweat_smile:

arnaldo2792 (approved):

Dockerfile
markusboehme: It's the kernel buildroot takes the kernel headers from, used e.g. for glibc and musl. Each buildroot release has some predefined versions and hashes for them that were current at the time. We need to take the earliest kernel still featured in any Bottlerocket release, but otherwise don't need to worry about updating since the kernel ABI won't/is not supposed to break. (One thread to pull on: https://git.busybox.net/buildroot/tree/package/linux-headers/Config.in.host?h=2022.02.5#n382)

patches/buildroot/0001-disable-shared-for-host-builds-of-gmp-isl-mpc-mpfr.patch
markusboehme: Ben likely passed `--no-numbered` to `git format-patch`. There was a discussion on a PR lately (that I can't find right now). The gist of it is that the patch numbers in the subject line don't add extra information and cause needless changes when adding or removing a patch, so we decided to try without them for a while and see whether we like that better.

Dockerfile
bcressey: The permissions get fixed up later in the copy to the final image. Here we need the directories to be owned by `builder` in order for the license scan tool to write the attribution files for the vendored dependencies.

hashes/go
```diff
@@ -1,2 +1,2 @@
-# https://golang.org/dl/go1.18.2.src.tar.gz
-SHA512 (go1.18.2.src.tar.gz) = 9214cbc051cf26b49ab1bd4d8d30b060865944b831578a9ea7fcfe33f84009f77932a39f2948efbfc412b151e4734a3bc1f8332f3bea11b35a912b0e23a4118f
+# https://go.dev/dl/go1.18.6.src.tar.gz
+SHA512 (go1.18.6.src.tar.gz) = 2af66b09bfe033b413eb7603a73a490319bf49fec0a2e20c40350e60b9ef35250a6dc8544c5fc67bd1ede55e242d056e7749f69ef500a38b1efe4b8f93078de3
```
jpmcb: Is there a reason we don't want Go 1.19.1?

From reading the 1.19 release notes, there isn't anything breaking and we'd get the benefit of consuming some compiler optimizations

configs/cargo-deny/clarify.toml
```diff
@@ -28,27 +37,33 @@ skip-files = [
 ]
 
 [clarify.crossbeam-channel]
-expression = "(MIT OR Apache-2.0) AND BSD-2-Clause AND CC-BY-3.0"
+expression = "(MIT OR Apache-2.0) AND BSD-2-Clause-FreeBSD"
```
markusboehme: How did you determine the change in license for this crate? I looked at the GitHub repo and couldn't deduce it, but very well may not have gone deep enough.

Dockerfile
```diff
@@ -285,16 +285,17 @@ RUN \
...
 
 WORKDIR /home/builder
 COPY ./hashes/openssl ./hashes
 RUN \
   sdk-fetch hashes && \
   rpm2cpio openssl-${OPENSSLVER}-${OPENSSLREV}.*.src.rpm | cpio -idmv && \
-  tar xf openssl-${OPENSSLVER}-hobbled.tar.gz && \
+  tar xf openssl-${OPENSSLVER}-hobbled.tar.xz && \
   mv openssl-${OPENSSLVER} openssl && \
+  grep -Fwl '<sys/sdt.h>' *.patch | xargs rm && \
```
markusboehme: I see this is about `0053-Add-SHA1-probes.patch`. It might be better to explicitly remove that patch to avoid false positives--we certainly don't need this particular patch, but others that happen to mention the header (e.g. just in their patch context) could be ones we want to take.

hashes/go
bcressey: > Is there a reason we don't want Go 1.19.1?

No reason, I just needed to draw the scope boundary somewhere and a Go update was bigger than I wanted to tackle. Same story with `GCC`.

I can look at adding it next time along with Rust 1.64.0, which will be out later this week.

jpmcb: > I just needed to draw the scope boundary somewhere and a Go update was bigger than I wanted to tackle

Super fair! Good to know! Thanks Ben!!

markusboehme (approved):
There have been new releases for LLVM and buildroot in the meantime. This just as a PSA. The line has to be drawn somewhere, of course, and a convergent state never lasts very long. :-)

configs/cargo-deny/clarify.toml
bcressey: Oh, in this case I took it from [clarify.toml](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/clarify.toml#L48) in the core repo. It looks like it hasn't been accurate since https://github.com/crossbeam-rs/crossbeam/commit/33e4c64a5 removed the other license statement from `LICENSE-THIRD-PARTY`.

bcressey:
> There have been new releases for LLVM and buildroot in the meantime. This just as a PSA. The line has to be drawn somewhere, of course, and a convergent state never lasts very long. :-)

Yup, there's about to be the Rust 1.64.0 release as well 😢

Dockerfile
bcressey: That makes sense. I wanted to avoid having to adjust this every time the upstream patch numbers changed, but that's not a big deal, and I agree that it's better than the alternative of silently deleting a patch we might care about.

configs/cargo-deny/clarify.toml
bcressey: Opened https://github.com/bottlerocket-os/bottlerocket/pull/2447 and https://github.com/bottlerocket-os/bottlerocket-update-operator/pull/250 to fix this elsewhere in the org.

bcressey:
Going to try the Go and Rust updates while I'm at it, since I need to rebuild the images for the other fixes. Putting this in draft until I can retest.

bcressey:
Tests look good so flipping this back to ready.

hashes/go
```diff
@@ -1,2 +1,2 @@
-# https://golang.org/dl/go1.18.2.src.tar.gz
-SHA512 (go1.18.2.src.tar.gz) = 9214cbc051cf26b49ab1bd4d8d30b060865944b831578a9ea7fcfe33f84009f77932a39f2948efbfc412b151e4734a3bc1f8332f3bea11b35a912b0e23a4118f
+# https://go.dev/dl/go1.19.1.src.tar.gz
```
jpmcb, line 1: Huzzah! 👏🏼 go 1.19!

jpmcb (approved):

jpculp (approved):

PR #87, opened 2023-01-10
update dependencies and tools

bcressey:
**Issue number:**
N/A

**Description of changes:**
Update dependencies and tools.

**Testing done:**
Built Bottlerocket variants for x86_64 and aarch64.

**Terms of contribution:**


stmcginnis (approved):
Unable to test myself, but based on described testing and looking over the changes, everything looks fine to me. I guess I'd give it some bonus confidence points based on the author too.

jpmcb (approved):
Built locally via `make` on both x86 and arm hosts 👍🏼

jpculp (commented):
Should we bump tokio in the license-tool?

yeazelm:
We should make sure to grab Rust 1.66.1 which will include a fix for a CVE in `cargo` [CVE-2022-46176](https://github.com/rust-lang/wg-security-response/tree/main/patches/CVE-2022-46176)

Dockerfile
```diff
@@ -1,21 +1,16 @@
...
-    createrepo_c e2fsprogs gdisk python3-jinja2 \
-    kpartx lz4 veritysetup dosfstools mtools squashfs-tools \
-    perl-FindBin perl-IPC-Cmd perl-open policycoreutils \
-    secilc qemu-img glib2-devel rpcgen erofs-utils jq ShellCheck \
-    json-c-devel libcurl-devel p11-kit-devel && \
-  dnf clean all && \
+    bc bison cmake cpio curl dnf-plugins-core dwarves elfutils-devel flex g++ \
+    gcc git gperf hostname intltool jq json-c-devel kmod libtool libcurl-devel \
+    meson openssl openssl-devel p11-kit-devel perl-ExtUtils-MakeMaker \
+    perl-FindBin perl-IPC-Cmd perl-open python rsync wget which && \
```
webern: I would probably put one per line and sort it so that changes to the list are discernible. That's a debatable style choice of course...

webern (approved):

Dockerfile
bcressey: Good point - this is definitely the right time to do that, so everything is easier to review going forward.

jpculp (approved):

bcressey:
I've started running into this error, which looks like https://github.com/awslabs/aws-c-common/issues/964:
```
> [sdk-final 34/35] RUN   mkdir .netscape &&   certutil -N --empty-password:
#192 2.713 Fatal error condition occurred in /home/builder/aws-sdk-cpp-src/crt/aws-crt-cpp/crt/aws-c-common/source/allocator.c:209: allocator != ((void *)0)
#192 2.713 Exiting Application
#192 2.713 No call stack information available
```

I'm going to revert the aws-sdk-cpp update in order to keep this moving forward. I tried going back to the previous version in this PR - 1.10.22 - but it's also failing now.

zmrow (approved):
:champagne:

Dockerfile
```diff
@@ -4,20 +4,45 @@ FROM public.ecr.aws/docker/library/fedora:37 as base
...
+    rsync \
+    wget \
+    which \
+  && \
   dnf config-manager --set-disabled \
-    fedora-modular updates-modular fedora-cisco-openh264 && \
-  dnf clean all && \
+    fedora-modular \
+    updates-modular \
+    fedora-cisco-openh264 \
```
arnaldo2792, line 44: Nit: these commands are reformatted here, but were added in the first commit of the series. Probably the diff will be smaller if the format doesn't change(?).

arnaldo2792 (approved):

webern (approved):

Dockerfile
stmcginnis: I think they should have been this newer format originally, and the sooner we change it then the sooner it's in better shape for ongoing updates where the diff can show individual lines more clearly.

arnaldo2792: Right, so probably keep this format in the first commit of the series

stmcginnis: This gets back to whether we treat our unit of change as individual commits or the pull request. I personally am not as concerned about individual commits in a PR unless there is a chance one of those commits may need to be reverted at some point.

jpculp: I'm not sure it would be worth the hassle.

bcressey: Fair point - I'd tidy this up if there were another change, but I don't think it's worth the retesting burden just for commit history. Style-wise, that was consistent with the rest of `Dockerfile` at the time the change was made, so it's not indefensible. 😀

PR #94, opened 2023-02-22
sdk-fetch: Allow upstream source fallback

jpculp:
**Description of changes:**

This change allows for using upstream sources as a fallback when **UPSTREAM_SOURCE_FALLBACK** is set to 'true'.

**Testing done:**

Ran `make` with a versioned file not available in the lookaside cache which failed (as expected), then ran again with `make UPSTREAM_SOURCE_FALLBACK=true` which downloaded successfully.

**Terms of contribution:**

### DIFF
--- a/sdk-fetch
+++ b/sdk-fetch
@@ -1,8 +1,13 @@
 #!/bin/bash
 set -euxo pipefail
+UPSTREAM_SOURCE_FALLBACK="${UPSTREAM_SOURCE_FALLBACK:-'false'}"
 # shellcheck disable=SC2046
-curl --fail --remote-name-all --remote-time \
+if ! curl --fail --remote-name-all --remote-time \
     $(awk -F '[ ()]' '/^SHA512 \(/ {
         printf "https://cache.bottlerocket.aws/%s/%s/%s\n", $3, $6, $3
-    }' "$1")
+    }' "$1") \
+&& [[ "${UPSTREAM_SOURCE_FALLBACK}" == 'true' ]]; then
+    curl --fail --remote-name-all --remote-time --location \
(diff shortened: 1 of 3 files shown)


sdk-fetch
```diff
@@ -1,8 +1,13 @@
...
 # shellcheck disable=SC2046
-curl --fail --remote-name-all --remote-time \
+if ! curl --fail --remote-name-all --remote-time \
     $(awk -F '[ ()]' '/^SHA512 \(/ {
         printf "https://cache.bottlerocket.aws/%s/%s/%s\n", $3, $6, $3
-    }' "$1")
+    }' "$1") \
+&& [[ "${UPSTREAM_SOURCE_FALLBACK}" == 'true' ]]; then
+    curl --fail --remote-name-all --remote-time --location \
+        $(awk '/^#\s((s)?ftp|http(s)?):\/\// {printf "%s\n", $2}' "$1")
```
jpculp, line 11: This could be simplified as `$(awk '/^# http/ {printf "%s\n", $2}' "$1")`, but the committed approach felt more complete.

jpmcb (approved):

stmcginnis (approved):
Looks good!

sdk-fetch
arnaldo2792: We are still in control of the checksums file right?

jpculp: Right, the checksums are still compared to the `hashes/my-thing` files. The only thing this does different is rather than curl the generated lookaside url it will curl whatever (s)ftp/http(s) urls are in the comment lines of those files.

arnaldo2792 (approved):

webern (approved):

jpculp:
Rebased.

PR #111, opened 2023-06-08
remove openssl build

bcressey:
**Issue number:**
N/A

**Description of changes:**
This build was used by the update operator and the test system, both of which have now switched to rustls.

**Testing done:**
Verified that the SDK built and the musl sys-root was back to just kernel headers, musl, and libunwind.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -326,73 +326,6 @@ RUN \
 
 # =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
-FROM sdk-musl as sdk-musl-openssl
-USER builder
-
-ARG OPENSSLVER="3.0.8"
-ARG OPENSSLREV="1"
-
-WORKDIR /home/builder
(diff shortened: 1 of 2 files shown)


cbgbt (approved):
:crab:

jpculp (approved):

jpmcb (approved):
🕺🏼

webern (changes requested):
This is just to remove the static build, right? I needed the regular fedora install of openssl for something during the twoliter prototyping. Should I look into what that was, or is it irrelevant for this PR?

bcressey:
> This is just to remove the static build, right? I needed the regular fedora install of openssl for something during the twoliter prototyping. Should I look into what that was, or is it irrelevant for this PR?

Right, the "host" openssl build from Fedora will still be present.

arnaldo2792 (approved):

webern (approved):
The openssl dependency I was thinking of occurs when `tuftool` generates a default root key in `pubsys-setup`. We will find another way to do this.

PR #114, opened 2023-06-14
Add short SHA to image tags

jpculp:
**Description of changes:**

This changes the default image tag to include an 8 character SHA of the commit that it was built off of.

*(ex. v0.33.0-abadc5c6-aarch64)*

Including the SHA will be a bit more in line with some other containers we build such as the bottlerocket-admin-container, but would also ensure that both arched builds in the manifest came from the same commit. The manifest tag itself won't include SHA and will be the same as what we publish today.

**Testing done:**

Built the SDK with `make ARCH=x86_64` on an `aarch64` host and a `x86_64` host, published with the `publish` target,  then pulled both platforms from the manifests in ECR.

**Terms of contribution:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,10 @@ HOST_ARCH ?= $(shell uname -m)
 UPSTREAM_SOURCE_FALLBACK ?= false
 
 VERSION := $(shell cat $(TOP)VERSION)
+SHORT_SHA := $(shell git rev-parse --short=8 HEAD)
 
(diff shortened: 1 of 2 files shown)


bcressey:
Curious why we'd do this - the description and commit message don't really explain it.

rpkelly (approved):

jpculp:
- Updated the description with additional context.
- Tweaked **SHORT_SHA** assignment to ensure lowercase.

bcressey (approved):

Issue #108, opened 2023-06-05
add software to the sdk to support twoliter


webern:
### Description

Add the following things to the `bottlerocket-sdk`:

- [x] `cargo make`
- [x] ~~`cargo deny`~~
- [x] `docker` (the CLI only, not the daemon)
- [x] Bottlerocket's RPM macros
- [x] Bootconfig data files

### Definition of Done

When the above items have been added to the SDK.

#### Depends on:

- [x] https://github.com/bottlerocket-os/twoliter/issues/6

```
id: texec_sdk
epic: twoliter_exec
```

Related to https://github.com/bottlerocket-os/bottlerocket/issues/2669

webern:
It's unusual for us to require something to be released before the GitHub issue is done. So I changed the definition of done and added https://github.com/bottlerocket-os/bottlerocket/issues/3187

bcressey:
We already have `cargo-deny` in the SDK, so I crossed that off.

bcressey:
For the bootconfig data, we have two files:
* `empty-bootconfig.data` - addressed in https://github.com/bottlerocket-os/bottlerocket/pull/3189
* `qemu-x86-console-bootconfig.data` - used by `start-local-vm`

I'm not sure what the fate of the other content in `tools` will be long-term; maybe it all gets relocated to a "contrib" repo of some sort, or rewritten as `twoliter` helper commands (e.g. `twoliter-start-local-vm` can be a local helper script invoked by `twoliter start-local-vm ...`), which can embed the data file as a base64 string or whatever.

webern:
> I'm not sure what the fate of the other content in `tools`

Most of the rest of the tools will go to the twoliter GitHub repository and be "embedded" in with the twoliter CLI. But you're right, I haven't thought about the "workflow" and "helper" type scripts. It's possible those conveniences could also be added to twoliter.

bcressey:
Added these to the SDK.

PR #115, opened 2023-06-16
merge macros from the monorepo

bcressey:
**Issue number:**
#108

**Description of changes:**
Add the macro-related commits from the Bottlerocket repo to a new root branch, and merge it into the SDK history.

Branch created as follows:
```
git clone git@github.com:bottlerocket-os/bottlerocket.git
cd bottlerocket
git checkout -b monorepo-macros
git fast-export monorepo-macros -- macros/* >../macros.fi
cd ..
git clone git@github.com:bcressey/bottlerocket-sdk.git
cd bottlerocket-sdk
git switch --orphan monorepo-macros
git fast-import < ../macros.fi
git checkout develop
git checkout -b merge-monorepo-macros
git merge --signoff --allow-unrelated-histories monorepo-macros
git push -u origin merge-monorepo-macros
```

**Testing done:**
Branch looks OK in my local git history.

**Terms of contribution:**

### DIFF
--- a/macros/aarch64
+++ b/macros/aarch64
@@ -1,0 +1,1 @@
+%_cross_arch aarch64
--- a/macros/aarch64
+++ b/macros/aarch64
@@ -1,4 +1,5 @@
 %_cross_arch aarch64
 %_cross_karch arm64
+%_cross_kimage Image
 %_cross_cpu_family arm
 %_cross_endian little
--- a/macros/aarch64
+++ b/macros/aarch64
@@ -3,3 +3,10 @@
 %_cross_kimage Image
 %_cross_cpu_family arm
 %_cross_endian little
+
+%_cross_grub_target aarch64
+%_cross_grub_platform efi
+%_cross_grub_tuple arm64-efi
+%_cross_grub_image bootaa64.efi
+%_cross_grub_prefix /EFI/BOOT
+%_cross_grubdir /boot/efi/EFI/BOOT/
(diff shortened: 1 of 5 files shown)


bcressey:
From a review standpoint, I'd appreciate feedback on the Git mechanics. I've verified that the macros end up the same, so there's not any new code to look at.

I have a few other changes I want to make to macros, and I need to actually integrate them into the SDK, but I'm holding off until this goes in.

cbgbt (approved):
Makes sense to me and the history looks correct, but I didn't realize that git had import/export prior to reading this review, so I'm not sure how much my thumbs up is worth :smile:

webern (approved):

markusboehme:
Cutting off history always is a bit unpleasant, but I'm happy we're retaining the logs. It should be reasonable easy to find related changes in the mono repo with that.

As for the Git mechanics, it's a nice use of the import functionality, too--much nicer than `filter-branch`! Having multiple roots like this in a repo may be unusual, but is fine. (For example, btrfs was brought into the Linux repo like that.)

The `macros` directory isn't changing too much, so keeping the mono^Wmain repository and the SDK repo synced is no big deal in the short term. The plan is to remove `macros` from the main repo after the next SDK release though, right?

markusboehme (approved):

foersleo (approved):

bcressey:
> The macros directory isn't changing too much, so keeping the mono^Wmain repository and the SDK repo synced is no big deal in the short term. The plan is to remove macros from the main repo after the next SDK release though, right?

Right!

PR #116, opened 2023-06-19
add additional arguments to `publish-sdk`

jpculp:
**Description of changes:**

Adds the ability to skip publishing a multi-arch manifest (`--skip-manifest`) or publish a multi-arch manifest with remote images that were pushed beforehand (`--only-manifest`).

Also adds a counter for how many ECR operations _(container/manifest pushes)_ occurred and compare it to the expected minimum number of operations. If the count is lower than expected _(ex. nothing got pushed)_, the script will return a 1.

**Testing done:**
- Built SDK
- Built SDK with SKIP_MANIFEST="true"
- Built SDK with ONLY_MANIFEST="true"

**Terms of contribution:**

### DIFF
--- a/publish-sdk
+++ b/publish-sdk
@@ -3,6 +3,9 @@
 set -e -o pipefail
 shopt -qs failglob
 
+SKIP_MANIFEST="${SKIP_MANIFEST:-'false'}"
+ONLY_MANIFEST="${ONLY_MANIFEST:-'false'}"
+
 for opt in "$@"; do
    optarg="$(expr "${opt}" : '[^=]*=\(.*\)')"
    case "${opt}" in
@@ -10,12 +13,18 @@ for opt in "$@"; do
       --registry=*) REGISTRY="${optarg}" ;;
       --sdk-name=*) SDK_NAME="${optarg}" ;;
       --short-sha=*) SHORT_SHA="${optarg,,}" ;;
+      --skip-manifest=*) SKIP_MANIFEST="${optarg,,}" ;;
+      --only-manifest=*) ONLY_MANIFEST="${optarg,,}" ;;
    esac
 done
 
 for required_arg in VERSION REGISTRY SDK_NAME SHORT_SHA; do
   [ -z "${!required_arg}" ] && echo "${required_arg} not set!" >&2 && exit 1
 done
+
+if [[ "${SKIP_MANIFEST}" == "true" ]] && [[ "${ONLY_MANIFEST}" == "true" ]] ; then
+  echo "nothing to do" >&2 && exit 2
+fi
 
 update_image() {
   set -u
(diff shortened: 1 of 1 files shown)


publish-sdk
```diff
@@ -10,13 +10,21 @@ for opt in "$@"; do
       --registry=*) REGISTRY="${optarg}" ;;
       --sdk-name=*) SDK_NAME="${optarg}" ;;
       --short-sha=*) SHORT_SHA="${optarg,,}" ;;
+      --skip-manifest=*) SKIP_MANIFEST="${optarg,,}" ;;
+      --only-manifest=*) ONLY_MANIFEST="${optarg,,}" ;;
```
bcressey, line 17: Can we error out if `SKIP_MANIFEST` and `ONLY_MANIFEST` are both set, or exit immediately with "nothing to do" in that case?

publish-sdk
```diff
@@ -51,64 +59,70 @@ update_image() {
...
   local remote_alt_arch_image="${remote_image}-${SHORT_SHA}-${alt_arch}"
 
-  if ! docker image inspect "${local_host_arch_image}" >/dev/null 2>&1 ; then
-    echo "did not find ${local_host_arch_image}, skipping it ..." >&2
-    return
-  fi
-
-  if ! docker tag "${local_host_arch_image}" "${remote_host_arch_image}" ; then
-    echo "failed to tag ${remote_host_arch_image}" >&2 && exit 1
+  if [[ "${ONLY_MANIFEST:-"false"}" != "true" ]]; then
```
bcressey: Might be good to add the `u` option to error on unset variables, then default both of the new ones to "false" at the top of the script.

publish-sdk
```diff
@@ -51,64 +59,70 @@ update_image() {
...
-
-  if ! docker manifest annotate "${remote_image}" \
-      "${remote_alt_arch_image}" --arch "${alt_docker_arch}" ; then
-    echo "failed to annotate manifest ${remote_image} with ${remote_alt_arch_image}" >&2 && exit 1
-  fi
-
-  # push the manifest and remove the local cache
-  if ! docker manifest push --purge "${remote_image}" ; then
-    echo "failed to push ${remote_image}" >&2 && exit 1
+  if [[ "${SKIP_MANIFEST:-"false"}" != "true" ]]; then
```
bcressey: You can invert the conditional and just return early if `SKIP_MANIFEST` is "true". That will avoid a lot of the indentation-related changes.

publish-sdk
jpculp: I had considered that, but I wanted to allow for **ONLY_MANIFEST** and **SKIP_MANIFEST** to be manipulated with environment variables. It's useful for automated build systems like AWS CodeBuild. I could probably still add the `u` option and just set `ONLY_MANIFEST="${ONLY_MANIFEST:-"false"}"` at the top for the best of both worlds.

jpculp: It looks like we already set `u` within the update image function so we can have the nicer required arguments error. I'll still go ahead and set those two variables at the top though.

jpculp:
- Default to "false" when **SKIP_MANIFEST** or **ONLY_MANIFEST** are unset.
- Exit early if both **SKIP_MANIFEST** and **ONLY_MANIFEST** are both "true".
- Return early if  **SKIP_MANIFEST** is "true".

bcressey (approved):

etungsten (approved):

PR #121, opened 2023-07-26
Update Rust to 1.71.0

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This bumps the rust toolchain to the latest 1.71.1 release.

Also adds a helper script to make this process easier and a little more automated for future updates.

**Testing done:**

Built aws-k8s-1.24 using updated SDK. Created cluster and ran `sonobuoy run --mode=quick` - all tests passed. `kubectl get nodes` shows everything healthy.

Built aws-k8s-1.27 using updated SDK. Created cluster and ran `sonobuoy run --mode=quick` - all tests passed. `kubectl get nodes` shows everything healthy. Checked `systemctl status kubelet` and `journalctl -u kubelet`, no errors seen and everything appears happy.

**Terms of contribution:**

### DIFF
--- a/tools/update-rust.sh
+++ b/tools/update-rust.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+set -e
+
+if [[ $# -lt 1 ]]; then
+    echo "Usage: $0 $RUST_VERSION"
+    echo
+    echo "Example: $0 1.71.0"
+    exit 2
+fi
+
+TOOLSDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOTDIR=$(realpath "${TOOLSDIR}/..")
+
+VERSION="${1}"
+OUTPUT="${ROOTDIR}/hashes/rust"
+PACKAGE_ROOT="https://static.rust-lang.org"
+METADATA_FILE="metadata.json"
+ARCHES=("x86_64" "aarch64")
+PACKAGES=("rust-std" "rustc" "cargo")
+
+function get_package_url {
+    local package="$1"
+    local arch="$2"
+
+    path=$(grep -e "${package}.*${arch}-unknown-linux-gnu.tar.xz" "${METADATA_FILE}" | cut -d '"' -f 2)
+
+    echo "${path}"
+}
+
+function add_package {
+    local package="$1"
+    local arch="$2"
+
+    # Lookup and pull down the package files
+    PKG_PATH=$(get_package_url "${package}" "${arch}")
+    PKG_FILE=$(echo "${PKG_PATH}" | cut -d '/' -f 3)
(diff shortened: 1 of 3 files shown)


tools/update-rust.sh
```diff
@@ -0,0 +1,74 @@
...
+    local package="$1"
+    local arch="$2"
+
+    path=$(grep -e "${package}.*${arch}-unknown-linux-gnu.tar.xz" "${METADATA_FILE}" | cut -d '"' -f 3)
+    path=$(echo "${path//\\}")
+
+    echo "${path}"
+}
+
+function add_package {
```
jpculp, line 31: Can we also check the signatures with the `*.tar.xz.asc` files?

tools/update-rust.sh
```diff
@@ -0,0 +1,74 @@
...
+    echo "SHA512 (${PKG_FILE}) = ${PKG_SHA}" >> "${OUTPUT}"
+}
+
+cd /tmp
+
+# Get the rustc source package
+RUSTC_PACKAGE="rustc-${VERSION}-src.tar.xz"
+RUSTC_URL="${PACKAGE_ROOT}/dist/${RUSTC_PACKAGE}"
+
+wget "${RUSTC_URL}"
```
jpculp: I'm admittedly a `wget` fan, but I might recommend using `curl` instead since it's more likely to be installed on everyone's machine.

tools/update-rust.sh
stmcginnis: Good idea - done!

tools/update-rust.sh
stmcginnis: Switched

jpculp (approved):

tools/update-rust.sh
```diff
@@ -0,0 +1,94 @@
...
+METADATA_FILE="metadata.json"
+ARCHES=("x86_64" "aarch64")
+PACKAGES=("rust-std" "rustc" "cargo")
+
+function get_package_url {
+    local package="$1"
+    local arch="$2"
+
+    path=$(grep -e "${package}.*${arch}-unknown-linux-gnu.tar.xz" "${METADATA_FILE}" | cut -d '"' -f 2)
+    path=$(echo "${path//\\}")
```
markusboehme: Why does the script need to strip backslashes? Nit: If it turns out to be needed, it could be just `path=${path//\\}`.

markusboehme (commented):
Changes look good to me and the script made verifying I arrive at the same changes a breeze. :+1: Approving once test results are in.

tools/update-rust.sh
stmcginnis: I think originally I was pulling this a different way that was resulting in a trailing `\`. But that is no longer the case, and I've verified this line is not needed. Will clean that up.

markusboehme (approved):
:racehorse:

PR #123, opened 2023-08-01
Revert "Update Go to 1.20.6"

stmcginnis:
**Issue number:**

Related: https://github.com/moby/moby/issues/45935

**Description of changes:**

This reverts the bump to Go 1.20.6 until the Moby issue can be investigated. It's possible there are no problems, but reading the bug report, there's a pretty fundamental change in this patch release. Original testing was a quick validation, since nothing was expected with a patch release, but this will require more complete testing with `k8s` and `ecs` variants to make sure there are no surprises.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -451,7 +451,7 @@ FROM sdk-libc as sdk-go
 
 ARG ARCH
 ARG TARGET="${ARCH}-bottlerocket-linux-gnu"
-ARG GOVER="1.20.6"
+ARG GOVER="1.20.5"
 
 USER root
 RUN dnf -y install golang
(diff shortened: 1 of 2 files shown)


markusboehme (approved):

stmcginnis:
Confirmed the way containers are launched via ECS is able to successfully create and run a container. But trying to directly use `docker` from the console results in:

```bash
bash-5.1# docker run -it --rm --name testing fedora bash
http: invalid Host header
```

Directly using `docker` isn't really a supported use case, but it can be a very useful troubleshooting tool and we should make sure it works.

So confirming that we should revert this change and work on getting a patch in place.

stmcginnis:
We can abandon this if https://github.com/bottlerocket-os/bottlerocket/pull/3312 is merged. Leaving open for now, but I think the fix is the right answer to allow us to move forward. Go 1.21.0 is now out, so we should keep up to date.

stmcginnis:
#3312 has merged, no need to revert this now.

PR #124, opened 2023-08-06
license-scan: allow clarification version reqs

### LINKED ISSUE #98: support multiple license file hashes in license-scan
When pulling multiple versions of a dependency that requires a clarify.toml it is possible to get in a situation where the license hash differs between the versions. Since we can only pass one hash per filename, the license check will always fail for one of the two. At a minimum, we should support multiple hashes for a given filename.

cbgbt:
**Description of changes:**
Closes #98 

```
license-scan: allow clarification version reqs

Clarifications can now optionally provide semver version specifiers.
See https://docs.rs/semver/latest/semver/enum.Op.html for more details
on allowed specifiers.
```

While trying to [update the `regex` crate](https://github.com/bottlerocket-os/bottlerocket-update-operator/pull/482) for the [bottlerocket-update-operator](https://github.com/bottlerocket-os/bottlerocket-update-operator), two major versions of `regex-automata` are pulled in due to [`matchers` being on an older version](https://github.com/hawkw/matchers/issues/3).

Unfortunately, `clarify.toml`'s existing clarification for `regex-automata-0.1` clashes with the new licensing for `regex-automata-0.3`. This adds the ability to clarify both versions.

**Limitations**
This currently assumes that the versions will follow semver; however, this implementation does not close the door to other version comparisons. If need be, we can replace `VersionReq` with an enum that also supports fallback to something like [RPM version comparison](https://fedoraproject.org/wiki/Archive:Tools/RPM/VersionComparison)

**Testing done:**
* Provided unit tests pass

**Terms of contribution:**

### DIFF
--- a/license-scan/Cargo.lock
+++ b/license-scan/Cargo.lock
@@ -63,6 +63,8 @@ dependencies = [
  "cargo_metadata",
  "ignore",
  "lazy_static",
+ "maplit",
+ "semver 1.0.18",
  "serde",
  "spdx",
  "structopt",
@@ -93,7 +95,7 @@ version = "0.11.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a3a567c24b86754d629addc2db89e340ac9398d07b5875efcff837e3878e17ec"
 dependencies = [
- "semver",
+ "semver 0.10.0",
  "serde",
  "serde_json",
 ]
@@ -292,6 +294,12 @@ checksum = "abb12e687cfb44aa40f41fc3978ef76448f9b6038cad6aef4259d3c095a2382e"
 dependencies = [
  "cfg-if",
 ]
+
+[[package]]
+name = "maplit"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3e2e65a1a2e43cfcb47a895c4c8b10d1f4a61097f9f254f183aee60cad9c651d"
 
 [[package]]
 name = "memchr"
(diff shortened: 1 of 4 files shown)


stmcginnis:
Nice! Looks like this could close https://github.com/bottlerocket-os/bottlerocket-sdk/issues/98

arnaldo2792 (approved):

cbgbt:
> Nice! Looks like this could close #98

@stmcginnis nice catch! I modified a test to make this more clear and updated the PR description.

jpculp (approved):

cbgbt:
I chatted with @jpculp that tought me that you have to explicitly run `clippy` against tests with `cargo clippy --tests`

Fixed a lint against the test module.

stmcginnis:
Ironically, this now fails due to multiple versions with `cargo-deny`:

```
error[B004]: found 2 duplicate entries for crate 'semver'
   ┌─ /home/stmcg/src/bottlerocket-os/bottlerocket-sdk/license-scan/Cargo.lock:58:1
   │  
58 │ ╭ semver 0.10.0 registry+https://github.com/rust-lang/crates.io-index
59 │ │ semver 1.0.18 registry+https://github.com/rust-lang/crates.io-index
   │ ╰───────────────────────────────────────────────────────────────────^ lock entries
   │  
   = semver v0.10.0
     └── cargo_metadata v0.11.4
         └── bottlerocket-license-scan v0.1.0
   = semver v1.0.18
     └── bottlerocket-license-scan v0.1.0

bans FAILED, licenses ok, sources ok
```

Submitted https://github.com/bottlerocket-os/bottlerocket-sdk/pull/126 to fix that.

We should probably look in to some basic GitHub Action checks to run for this repo...

cbgbt:
😵

> We should probably look in to some basic GitHub Action checks to run for this repo...

I should have also just run the project-level build, rather than assume that this crate's testing would be sufficient. I forgot about the additional top-level checks. My bad!

PR #133, opened 2023-09-07
Add helper script to update go version

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

Similar to the recently added update-rust script, this takes the version of Go desired and fetches the file to verify and calculate the hash. It then updates the `hashes/go` file to use the requested version.

**Testing done:**

Used script to pull various releases and verified it works as expected.

**Terms of contribution:**

### DIFF
--- a/tools/update-go.sh
+++ b/tools/update-go.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [[ $# -lt 1 ]]; then
+    echo "Usage: $0 $GO_VERSION"
+    echo
+    echo "Example: $0 1.21.1"
+    exit 2
+fi
+
+TOOLSDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOTDIR=$(realpath "${TOOLSDIR}/..")
+
+VERSION="${1}"
+OUTPUT="${ROOTDIR}/hashes/go"
+PACKAGE_ROOT="https://go.dev/dl"
+
+# Get the go source package
+# e.g. https://go.dev/dl/go1.21.1.src.tar.gz
+GO_SRC_PACKAGE="go${VERSION}.src.tar.gz"
+GO_SRC_URL="${PACKAGE_ROOT}/${GO_SRC_PACKAGE}"
+
+curl -s -L -O -C - "${GO_SRC_URL}"
+
+# SHA256 is only used to validate the package with the published checksum
+GO_256_SHA=$(sha256sum "${GO_SRC_PACKAGE}" | cut -d ' ' -f 1)
+
+# The Go downloads are not signed, and there is not an easy checksum file to
+# fetch to validate the download. This is a little fragile (will need to update
+# if they change the download page structure) but it at least gives some level
+# of validation that we have a good download.
+PUBLISHED_MATCH=$(curl -s https://go.dev/dl/ | grep -B5 "${GO_256_SHA}" | grep "${GO_SRC_PACKAGE}")
+if [[ -z "${PUBLISHED_MATCH}" ]]; then
+    echo "Unable to verify source package checksum!!!"
+    echo
+    echo "Check ${GO_SRC_PACKAGE} contents compared to ${PACKAGE_ROOT}"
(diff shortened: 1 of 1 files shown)


tools/update-go.sh
```diff
@@ -0,0 +1,52 @@
...
+curl -s -L -O -C - "${GO_SRC_URL}"
+
+# SHA256 is only used to validate the package with the published checksum
+GO_256_SHA=$(sha256sum "${GO_SRC_PACKAGE}" | cut -d ' ' -f 1)
+
+# The Go downloads are not signed, and there is not an easy checksum file to
+# fetch to validate the download. This is a little fragile (will need to update
+# if they change the download page structure) but it at least gives some level
+# of validation that we have a good download.
+PUBLISHED_MATCH=$(curl -s https://go.dev/dl/ | grep -B5 "${GO_256_SHA}" | grep "${GO_SRC_PACKAGE}")
```
markusboehme, line 33: A somewhat more robust way might be to just check for the presence of the hashsum on the page (skipping the second command in the pipe) and just assuming that if there's a match, we got the right file and we got all of it. I don't mind either way, though--I doubt the website will change a lot.

markusboehme (approved):

tools/update-go.sh
stmcginnis: I thought about that, but figured there might be more risk if somehow the wrong thing got picked up. It would certainly make it simpler and more flexible. But I liked the slightly better validation of matching the hash with the expected package (without going too crazy and trying to actually parse HTML).

Maybe keep that as a fallback plan in case they actually do change the website? I do think it's pretty low risk. As long as I've been working with Go, I have not noticed this download link page change in any major way.

tools/update-go.sh
```diff
@@ -0,0 +1,52 @@
...
+GO_512_SHA=$(sha512sum "${GO_SRC_PACKAGE}" | cut -d ' ' -f 1)
+
+# Add the root/header information
+echo "# ${GO_SRC_URL}" > "${OUTPUT}"
+echo "SHA512 (${GO_SRC_PACKAGE}) = ${GO_512_SHA}" >> "${OUTPUT}"
+
+echo "================================================"
+echo "go toolchain updated to ${VERSION}"
+echo
+echo "Make sure to bump GOVER in Dockerfile to match"
```
yeazelm, line 51: Would it make sense to have the script also attempt to replace this line so that the user running the script only has to verify everything worked as expected vs do a mix of manual and automated work?

tools/update-go.sh
```diff
@@ -0,0 +1,52 @@
...
+VERSION="${1}"
+OUTPUT="${ROOTDIR}/hashes/go"
+PACKAGE_ROOT="https://go.dev/dl"
+
+# Get the go source package
+# e.g. https://go.dev/dl/go1.21.1.src.tar.gz
+GO_SRC_PACKAGE="go${VERSION}.src.tar.gz"
+GO_SRC_URL="${PACKAGE_ROOT}/${GO_SRC_PACKAGE}"
+
+curl -s -L -O -C - "${GO_SRC_URL}"
```
yeazelm, line 24: nit: It would be nice to download this to a temporary directory so it doesn't leave the tar file somewhere on the user's system to clean up since it won't be used directly in the build.

jpculp: I kind of prefer it _not_ being in a temporary directory, just to make it easier to copy to our cache.

jpculp (approved):

tools/update-go.sh
yeazelm: That is a good point. As a rule I tend to get annoyed at tools that "do something nice for me but don't clean up their mess" but in this case, its not a mess, its a useful artifact for the next step in the process.

stmcginnis: IIRC, on another of these I changed it. I was originally downloading to `/tmp`, but on Patrick's suggestion I changed it to be local specifically to help with that next step. So I do generally agree and also don't like things being left around, but in this case there is some value to it.

tools/update-go.sh
stmcginnis: Will look at this in a follow up. There are some other hash update things, so I'd like to take a look at all of it.

PR #147, opened 2023-11-27
update base image to Fedora 39

jpculp:
**Description of changes:**

* Fedora: **37 → 39**
* AWS SDK for C++: **1.9.332 → 1.11.207**

Updates the base image to the latest stable **Fedora**. Due to a newer version of `curl`, the **AWS SDK for C++** was bumped as well.

Modular repositories are no longer explicitly disabled as they were removed entirely in **Fedora 39**.

**Testing done:**

Built Bottlerocket.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -784,14 +784,16 @@ RUN \
 
 FROM sdk as sdk-cpp
 
-ARG AWS_SDK_CPP_VER="1.9.332"
+ARG AWS_SDK_CPP_VER="1.11.207"
 
 USER builder
 WORKDIR /home/builder/aws-sdk-cpp-src
 COPY ./hashes/aws-sdk-cpp /home/builder/aws-sdk-cpp-src/hashes
 
+# Upstream source fallback is explicitly disabled here as the SHA512 hash
+# verification fails due to a difference in the upstream names and the SDK's.
 RUN \
-  sdk-fetch hashes && \
+  UPSTREAM_SOURCE_FALLBACK=false sdk-fetch hashes && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -784,14 +782,16 @@ RUN \
...
 
 USER builder
 WORKDIR /home/builder/aws-sdk-cpp-src
 COPY ./hashes/aws-sdk-cpp /home/builder/aws-sdk-cpp-src/hashes
 
+# Upstream source fallback is explicitly disabled here as the SHA512 hash
+# verification fails due to a difference in the upstream names and the SDK's.
 RUN \
-  sdk-fetch hashes && \
+  UPSTREAM_SOURCE_FALLBACK=false sdk-fetch hashes && \
```
bcressey, line 794: Is the issue here that our files are named `aws-c-sdkutils.tar.gz` instead of `aws-c-sdkutils-a6fd80cf7c163062d31abb28f309e47330fbfc17.tar.gz`?

bcressey (approved):

Dockerfile
jpculp: Actually what will end up happening is the file will download as `a6fd80cf7c163062d31abb28f309e47330fbfc17`.

jpculp:
While building Bottlerocket we found that we had to add `--disable-nls` to the **procps** package. I'll add a commit for that in the same PR where we bump the SDK in Bottlerocket.

The **shim** package also flaked out in one of the test builds, but I was unable to get it to happen again. @bcressey thought it might be related to `%make_build` evaluating to `make -j$(nproc)`.

bcressey (approved):

etungsten (approved):

PR #156, opened 2024-03-04
include toolchains for both targets in one SDK

### LINKED ISSUE #137: eliminate toolchain container
**Image I'm using:**
0.34.1

**Issue or Feature Request:**
To simplify `twoliter` and its logic for handling SDK images, I'd like to eliminate the separate "toolchain" container and just have the regular SDK come with a toolchain archive for the eventual host architecture.

For example:
* x86_64 / aarch64 - should build a native aarch64 toolchain
* x86_64 / x86_64 - could either build a native toolchain or reuse the cross toolchain
* aarch64 / x86_64 - should also build a native x86_64 toolchain
* aarch64 / aarch64 - could either build a native toolchain or reuse the cross toolchain

This is [not supported by Buildroot](https://buildroot.org/downloads/manual/manual.html#faq-no-compiler-on-target) which is why the current implementation exists.

bcressey:
**Issue number:**
Fixes #137

**Description of changes:**
Build C, C++, and Rust toolchains for both supported targets - x86_64, aarch64 - into the same SDK image rather than one image per target.

Target-specific SDK images have a long history in Bottlerocket, dating back to the earliest days of the project where the toolchain was built alongside the rest of the packages. Ultimately it stems from our use of GCC and binutils, which do not natively support compiling for multiple target architectures the way that LLVM, Rust, and Go do.

When porting the SDK build logic to the Dockerfile, I kept the same basic structure where one stage corresponds to one of the previous packages. The stages were arch-specific just like the packages were. When adding support for toolchain archives to support kmod kits, I added a set of stages to collect just the C/C++ toolchains, and left it to downstream build logic to copy the right toolchain for a given architecture.

This change factors out the C/C++ build logic into helper scripts, and arranges to run those scripts once per architecture. The results are then merged back into the same stages that later stages depend on, with the main difference that both architectures are now present.

As a result, we no longer need either a separate toolchain image or arch-specific SDKs. The release footprint goes from 8 images - 4 SDK target/host arch combos, 4 toolchain target/host arch combos - down to 2: one SDK image per host arch.

There are a few compatibility breaks that need to be sorted out in `twoliter`. I've also opted to break the `make publish` interface by renaming the expected arguments, since the destination registry will change as a result of the refactoring done in `publish-sdk` and I don't want anyone to be surprised.

The SDK size is significantly increased, from 3 GiB to 4.5 GiB. This is unavoidable: we simply include more binaries now. However, in actual use this is partly mitigated in two ways. First, we no longer need separate toolchain images, saving 300 MiB per target arch used. Second, in contexts like CI where we use the SDK as a source of known-good, hash-pinned toolchains, we can end up pulling two copies of the SDK if the host and target architecture don't match. Now, the SDK is larger, but we only ever need to pull the one image.

**Testing done:**
Verified that only the two `ARG` directives that the `Makefile` uses are still present:
```
❯ rg '^ARG ' Dockerfile
85:ARG UPSTREAM_SOURCE_FALLBACK
136:ARG UPSTREAM_SOURCE_FALLBACK
340:ARG HOST_ARCH
895:ARG HOST_ARCH
```

Used the new `make` / `make publish` workflow to push images and generate a multi-arch manifest.
```
❯ docker image ls
REPOSITORY                                   TAG                        IMAGE ID       CREATED        SIZE
public.ecr.aws/c2b5m1s5/thar-be-a-beta-sdk   v0.40.0                    68318c32fbce   19 hours ago   4.65GB
public.ecr.aws/c2b5m1s5/thar-be-a-beta-sdk   v0.40.0-e48117fd-aarch64   68318c32fbce   19 hours ago   4.65GB
```

Built Bottlerocket images on x86_64 and aarch64 hosts, for both target architectures. Verified that they came up, and that the expected architecture's binaries were built.

**Terms of contribution:**


Dockerfile
```diff
@@ -391,24 +383,28 @@ RUN \
...
-    sed -i -e '/("'${ARCH}-unknown-linux-${libc}'", .*),/a("'${ARCH}-${VENDOR}-linux-${libc}'", '${ARCH}_${VENDOR}_linux_${libc}'),' \
-      compiler/rustc_target/src/spec/mod.rs ; \
+  for arch in x86_64 aarch64 ; do \
+    for libc in gnu musl ; do \
+      cp compiler/rustc_target/src/spec/targets/${arch}_{unknown,${VENDOR}}_linux_${libc}.rs && \
+      sed -i -e '/let mut base = base::linux_'${libc}'::opts();/a base.vendor = "'${VENDOR}'".into();' \
+        compiler/rustc_target/src/spec/targets/${arch}_${VENDOR}_linux_${libc}.rs && \
+      sed -i -e '/ \.\.base::linux_'${libc}'::opts()/i vendor: "'${VENDOR}'".into(),' \
+        compiler/rustc_target/src/spec/targets/${arch}_${VENDOR}_linux_${libc}.rs && \
+      sed -i -e '/("'${arch}-unknown-linux-${libc}'", .*),/a("'${arch}-${VENDOR}-linux-${libc}'", '${arch}_${VENDOR}_linux_${libc}'),' \
```
webern, line 393: Are these effectively patches to support a "vendor" of Bottlerocket? Is declaring oneself a vendor not a supported build argument?

jpculp (approved):

helpers/glibc/build-glibc.sh
```diff
@@ -0,0 +1,45 @@
...
+  --disable-multi-arch \
+  --disable-profile \
+  --disable-systemtap \
+  --disable-timezone-tools \
+  --disable-tunables \
+  --without-cvs \
+  --without-gd \
+  --without-selinux
+
+make -j$(nproc) -O -r
```
webern: Doesn't it automatically use the right number of procs?

webern (approved):
:sunglasses:

Dockerfile
```diff
@@ -961,13 +961,16 @@ RUN \
...
 
 FROM sdk as sdk-macros
-ARG ARCH
 
 COPY macros/* /tmp/
 
 WORKDIR /tmp
 RUN \
-  cat ${ARCH} shared rust cargo > /etc/rpm/macros.bottlerocket
+  for arch in x86_64 aarch64 ; do \
```
arnaldo2792, line 969: Will this change require rework on how the packages are compiled in Bottlerocket?

Dockerfile
```diff
@@ -443,7 +443,7 @@ RUN \
 
 FROM sdk-libc as sdk-go
 
-ENV GOVER="1.21.5"
+ENV GOVER="1.21.7"
```
arnaldo2792: They just released a newer version with security fixes:

https://github.com/golang/go/issues?q=milestone%3AGo1.21.8+label%3ACherryPickApproved

VERSION
```diff
@@ -1 +1 @@
-v0.37.0
+v0.40.0
```
arnaldo2792, line 1: I'm curious, why did you skip 2 versions?

Dockerfile
jpculp: Correct. I went ahead and opened/merged https://github.com/bottlerocket-os/bottlerocket-sdk/pull/158 so this could rebase on it.

VERSION
bcressey: > I'm curious, why did you skip 2 versions?

In case we ended up needing to update the previous style of SDK for a point release branch or something.

Dockerfile
bcressey: > Will this change require rework on how the packages are compiled in Bottlerocket?

It's transparent at the repo level but does require the supporting changes in `twoliter`.

helpers/glibc/build-glibc.sh
bcressey: > Doesn't it automatically use the right number of procs?

No, it's not automatic. You can use `-j` by itself to allow an unlimited number of jobs, but it's conventional to limit it to the number of CPUs.

Dockerfile
bcressey: > Are these effectively patches to support a "vendor" of Bottlerocket? Is declaring oneself a vendor not a supported build argument?

Yes - per the comment above, we need `*-bottlerocket-linux-*` targets to distinguish from `*-unknown-linux-*` targets when compiling for a target that matches the host architecture. The "vendor" is part of the target triple; you can override target on the command line but not the individual components like vendor or os.

It's possible to create a [custom target](https://docs.rust-embedded.org/embedonomicon/custom-target.html) using nightly cargo. Even when that stabilizes, I expect I'll still prefer to do it here in the SDK so that we don't have to build std and core "on demand" when building packages.

bcressey:
⬆️ force push includes rebase and some shellcheck fixes that @jpculp mentioned.

helpers/buildroot/build-gnu-toolchain.sh
```diff
@@ -0,0 +1,32 @@
...
+      --kernel-version=*) KVER="${optarg}" ;;
+   esac
+done
+
+ARCH="${ARCH:?}"
+KVER="${KVER:?}"
+
+cd "${HOME}/buildroot"
+make O="output/${ARCH}-gnu defconfig" BR2_DEFCONFIG="configs/sdk_${ARCH}_gnu_defconfig"
+make O="output/${ARCH}-gnu toolchain"
```
bcressey: I messed up the quoting here - will push a fix once I've tested the rest of the build.

arnaldo2792 (approved):

PR #162, opened 2024-03-21
support FIPS for Go builds

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket/issues/1667

**Description of changes:**
Conceptually this patch series has two halves.

The first half consists of patching Go, patching AWS-LC, building AWS-LC sysos, then building Go with those sysos instead of the stock "boringcrypto" ones. All of this is done so that `GOEXPERIMENT=boringcrypto go build ...` does the right thing. This extends the earlier unified SDK work in that we're able to replace the sysos for both amd64 and arm64, so the Go toolchain should just work for both architectures whether or not the "boringcrypto" experiment is enabled.

The second half makes it easier to use `GOEXPERIMENT=boringcrypto go build ...` in downstream packaging of Go binaries. There are new macros, including a new way to set up the environment for static builds, and a lint script that checks whether all Go binaries that should be are built in a parallel version with FIPS enabled. That script is currently disabled by default: it can be enabled on a per-package basis for testing, and will be enabled by default in a future SDK release, after packaging has a chance to catch up.

These changes should be backwards compatible with existing packaging - the AWS-LC bits are only used when `GOEXPERIMENT=boringcrypto` is set, the new lint doesn't run by default, and the old `%set_cross_go_flags` is mostly unchanged.

**Testing done:**
Made all the packaging changes required with a slightly different build where the lint script was on by default. (It helped me catch a ton of problems!) See https://github.com/bcressey/bottlerocket/commits/fips-and-non-fips/ for what this looks like in action with the (forthcoming) twoliter changes as well.

Built and launched `aws-k8s-1.28` AMIs where only the FIPS versions of Go binaries were used. The nodes came up, joined the cluster, ran pods, etc.

**Terms of contribution:**

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -17,3 +17,11 @@ operating system images.
 macros/rust and macros/cargo (used during build) are derived from the Fedora Rust SIG's rust2rpm.
 https://pagure.io/fedora-rust/rust2rpm
 Copyright (c) 2017 Igor Gnatenko. Licensed under the MIT License.
+
+=^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+helpers/go/parse-functions.go, helpers/go/umod-arm64.c, and helpers/go/umod-amd64.s are derived
+from the Go project's build-goboring.sh:
+https://github.com/golang/go/blob/master/src/crypto/internal/boring/build-goboring.sh
+Copyright (c) 2020 The Go Authors. Licensed under a BSD-3-Clause license.
+See helpers/aws-lc/LICENSE.
--- a/helpers/aws-lc/LICENSE
+++ b/helpers/aws-lc/LICENSE
@@ -1,0 +1,28 @@
+Copyright (c) 2009 The Go Authors. All rights reserved.
+Copyright Amazon.com, Inc., its affiliates, or other contributors.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+   * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
--- a/helpers/aws-lc/parse-functions.go
+++ b/helpers/aws-lc/parse-functions.go
@@ -1,0 +1,105 @@
+// Adapted from the Go project's build-goboring.sh:
+//   https://github.com/golang/go/blob/master/src/crypto/internal/boring/build-goboring.sh
+
+// Original Copyright 2022 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Modifications Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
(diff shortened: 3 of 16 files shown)


bcressey:
⬆️ force push to fix the FIXME

helpers/aws-lc/parse-functions.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// Locate the functions from the header file that we are interested in.
+	syms, err := locateFunctions(strings.Split(string(data), "\n"))
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	// Add extra functions that aren't in the header.
+	syms = append(syms, "BORINGSSL_bcm_power_on_self_test")
```
monirul, line 74: Why do we need this extra function? Can we add the reason/intention of adding this`BORINGSSL_bcm_power_on_self_test` in the comment?

helpers/aws-lc/parse-functions.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return syms, globals, renames, nil
+}
+
+func generateGlobalsRenames(syms []string) ([]string, []string) {
+	globals := make([]string, 0, len(syms))
+	renames := make([]string, 0, len(syms)+2)
+	for _, sym := range syms {
+		globals = append(globals, "_goboringcrypto_"+sym)
+		renames = append(renames, sym+" _goboringcrypto_"+sym)
```
monirul, line 85: Any reason for using `sym+" _goboringcrypto_"+sym` format?

helpers/aws-lc/parse-functions.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// Add the functions that we manually compile in.
+	renames = append(renames, "__umodti3 _goboringcrypto___umodti3", "__udivti3 _goboringcrypto___udivti3")
```
monirul, line 89: Why do we need this manual entry? Why those functions were manually compiled in? Should we elaborate the comment a bit more to document the reason?

helpers/aws-lc/parse-functions.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	// Write out our renames file, ensuring it ends with a newline to prevent some unix tools from failing.
+	if err := osWriteFile(filepath.Join(outputDir, "renames.txt"), []byte(strings.Join(renames, "\n")+"\n"), 0644); err != nil {
+		return fmt.Errorf("failed writing renames.txt: %w", err)
+	}
+
+	return nil
+}
+
+func parse(headerFile string) ([]string, []string, []string, error) {
```
monirul, line 60: Should we add unit tests for these functions to ensure changing in this file in future won't introduce any regression?

monirul (commented):
@bcressey , I added few comments, mostly related to the go code, as I'm trying to grasp the underlying intention of the code. Thanks.

helpers/aws-lc/parse-functions.go
bcressey: Overall I think the comment at the top suffices; this Go file is just meant as a more readable version of [build-goboring.sh](https://github.com/golang/go/blob/master/src/crypto/internal/boring/build-goboring.sh). It needs to be  understood in that context and track that source as the canonical reference for "how to integrate the libcrypto syso".

I didn't write this code (per commit message) and I'm not inclined to make any deeper investments in it. If it's a blocker then I'll just drop this and integrate the upstream awk scripts instead (somehow). I don't mind awk so much but figured the Go version would be more approachable to most reviewers.

monirul: Got it. 
>figured the Go version would be more approachable to most reviewers.

I agree. No need to change to awk script. I will mark the comments as resolved.

helpers/aws-lc/parse-functions.go
monirul: based on https://github.com/bottlerocket-os/bottlerocket-sdk/pull/162#discussion_r1534725834 reply, resolving it.

helpers/aws-lc/parse-functions.go
monirul: based on https://github.com/bottlerocket-os/bottlerocket-sdk/pull/162#discussion_r1534725834 reply, resolving it.

rpkelly (approved):

webern (approved):

macros/shared
```diff
@@ -239,8 +241,21 @@ CROSS_COMPILATION_CONF_EOF\
 %_source_payload w1.zstdio
 %_binary_payload w1.zstdio
 
+# Invocation for the check-fips buildroot policy script.
+%__cross_check_fips_cmd %{shrink: \
```
arnaldo2792, line 259: Shouldn't this be `__cross_check_fips_go_cmd`, since we may use/declare a different check in Rust binaries, or are you planning to do something different for those programs?

arnaldo2792 (approved):

macros/shared
bcressey: > Shouldn't this be __cross_check_fips_go_cmd, since we may use/declare a different check in Rust binaries, or are you planning to do something different for those programs?

I'd like to handle both Rust and Go in the same script if possible. If not, I can extend this macro to run both scripts.

helpers/aws-lc/LICENSE
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2009 The Go Authors. All rights reserved.
+
```
hyandell, line 3: Keeping this file as LICENSE makes sense given the references in the copied code; but add the project's copyright here too:

`Copyright Amazon.com, Inc., its affiliates, or other contributors.`

Given the scope of this code, having our changes be under BSD-3-Clause rather than Apache-2.0/MIT makes sense.

COPYRIGHT
```diff
@@ -17,3 +17,10 @@ operating system images.
 macros/rust and macros/cargo (used during build) are derived from the Fedora Rust SIG's rust2rpm.
 https://pagure.io/fedora-rust/rust2rpm
 Copyright (c) 2017 Igor Gnatenko. Licensed under the MIT License.
+
+=^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+helpers/go/parse-functions.go, helpers/go/umod-arm64.c, and helpers/go/umod-amd64.s are derived
+from the Go project's build-goboring.sh:
+https://github.com/golang/go/blob/master/src/crypto/internal/boring/build-goboring.sh
+Copyright (c) 2020 The Go Authors. Licensed under a BSD 3-Clause license.
```
hyandell: Change 'BSD 3-Clause' to 'BSD-3-Clause'. (i.e. use SPDX ids)

Add final sentence of:

See helpers/aws-lc/LICENSE.

hyandell (approved):

bcressey:
⬆️ force push to rebase and fix the `Dockerfile` conflict.

jpculp (approved):

PR #183, opened 2024-05-28
build: add GRUB for grub-bios-setup

### LINKED ISSUE #182: grub-bios-setup should be included in the SDK
**Image I'm using:**
v0.41.0

**Issue or Feature Request:**
Currently the GRUB package builds the install tools for [the host architecture](https://github.com/bottlerocket-os/bottlerocket/blob/0d6b452690a1944f1362be7ec90a83a90753a7a3/packages/grub/grub.spec#L145). Then `rpm2img` runs `grub-bios-setup` using [that binary](https://github.com/bottlerocket-os/twoliter/blob/cbd0bd9afdcedf6cd71884cd90ce1a5bb071adf6/twoliter/embedded/rpm2img#L235).

This will break when GRUB is moved into a kit, because the build host that built the kit might not be the same architecture as the one where the kit is ultimately used to create an image.

To resolve this, we should include `grub-bios-setup` built for the host architecture in the SDK.

bcressey:
**Issue number:**
Closes #182 

**Description of changes:**
For installing the GRUB BIOS image, we rely on a patched version of `grub-bios-setup` that skips the guesswork around finding the root device. This is a tool that runs on the host, and must be built for the host's architecture. Adapt the current GRUB build from the main repo, adding just the patch we need and copying out only the required tool.

**Testing done:**
Modified `twoliter` to use the SDK's `grub-bios-setup`. Verified that I could build an x86_64 image and boot on a system using BIOS firmware.

**Terms of contribution:**


jpculp (approved):

Dockerfile
```diff
@@ -415,6 +415,85 @@ ENV PATH="/usr/libexec/rust/bin:$PATH" LD_LIBRARY_PATH="/usr/libexec/rust/lib"
...
+USER builder
+WORKDIR /home/builder
+COPY ./hashes/grub /home/builder/hashes
+COPY ./patches/grub /home/builder/patches
+
+# This rather elaborate way of unpacking the sources and applying the
+# patches mimics the way GRUB is built in Bottlerocket, which in turn
+# mimics the way that Amazon Linux and Fedora build it.
+RUN \
+  sdk-fetch /home/builder/hashes && \
```
arnaldo2792, line 437: would it be worth having this as a script instead of a long  "one-liner"?

arnaldo2792 (approved):

webern (approved):

Dockerfile
bcressey: > Would it be worth having this as a script instead of a long "one-liner"?

The line I am drawing in my head:
* for builds that we only do once for the host arch, it's more idiomatic to do them inline in the Dockerfile
* for builds that we need to do for multiple versions or target architectures, it's better to do them in a script

At some point I could see converging on the second approach for everything, but it does make the Dockerfile more abstract and less useful as a way to immediately see how something is built.

PR #201, opened 2024-08-19
go: remove `GOROOT_FINAL`

### LINKED ISSUE #200: `GOROOT_FINAL` no longer respected in Go 1.23+
In the release notes https://go.dev/doc/go1.23

> Setting the GOROOT_FINAL environment variable no longer has an effect ([#62047](https://go.dev/issue/62047)). Distributions that install the go command to a location other than $GOROOT/bin/go should install a symlink instead of relocating or copying the go binary.

bottlerocket-sdk sets that here: https://github.com/bottlerocket-os/bottlerocket-sdk/blob/b21e50c7f1318d44e44b8e1fca37eec90ccd273b/helpers/go/build-go.sh#L15

The sdk base image, `public.ecr.aws/docker/library/fedora:39` is currently on Go 1.22.5, so this is not seemingly an issue (as seen in #199) but there's no telling when they may update to Go 1.23.x and affect the sdk.

ginglis13:
**Issue number:**

Closes: #200 

**Description of changes:**

`GOROOT_FINAL` is not supported in Go 1.23. the bottlerocket-sdk wasn't using it anyway :) In the sdk-final stage of the Dockerfile, the Go sources and build artifacts for respective versions are copied over directly from `/home/builder/sdk-go directory` in a Go version layer's respective directory:

https://github.com/bottlerocket-os/bottlerocket-sdk/blob/6b0525c30cd510713d4a18de484370844dfbc1c9/Dockerfile#L1239-L1261

This commit also disables a shellcheck warning that thinks `GO111MODULE` is unused in `helpers/go/build-go.sh`

**Testing done:**

```
$ make
$ docker run --rm bottlerocket-sdk:v0.43.0-cec1e119-arm64 ls /usr/libexec | grep go
go-1.22
go-1.23
$ docker run --rm -e GO_MAJOR=1.23 bottlerocket-sdk:v0.43.0-cec1e119-arm64 go version
go version go1.23.0 linux/arm64
$ docker run --rm -e GO_MAJOR=1.22 bottlerocket-sdk:v0.43.0-cec1e119-arm64 go version
go version go1.22.6 linux/arm64
```

**Terms of contribution:**

### DIFF
--- a/helpers/go/build-go.sh
+++ b/helpers/go/build-go.sh
@@ -12,12 +12,12 @@ done
 
 GOVER="${GOVER:?}"
 
-export GOROOT_FINAL="/usr/libexec/go-${GOVER}"
 cd "${HOME}/sdk-go/src"
 ./all.bash
 
 # Install the Go standard library and toolchain.
 cd "${HOME}/sdk-go"
+# shellcheck disable=SC2034
 PATH="${HOME}/sdk-go/bin:${PATH}" GO111MODULE="auto"
 go install -buildmode=pie std cmd
 install -p -m 0644 -Dt licenses LICENSE PATENTS


jpculp (approved):

helpers/go/build-go.sh
```diff
@@ -12,12 +12,11 @@ done
...
 GOVER="${GOVER:?}"
 
-export GOROOT_FINAL="/usr/libexec/go-${GOVER}"
 cd "${HOME}/sdk-go/src"
 ./all.bash
 
 # Install the Go standard library and toolchain.
 cd "${HOME}/sdk-go"
-PATH="${HOME}/sdk-go/bin:${PATH}" GO111MODULE="auto"
+export PATH="${HOME}/sdk-go/bin:${PATH}" GO111MODULE="auto"
```
bcressey: Why export these? They'll only be available to child processes of the shell, and the only one we care about is the `go` process that runs on the next line.

ginglis13: Hm yea you're right, I misinterpreted a shellcheck that `GO111MODULE` is unused, I'll revert and add a `# shellcheck disable=SC2034`

ginglis13:
^ force push reverts exporting `PATH` and `GO111MODULE` in `helpers/go/build-go.sh` and instead disables a shellcheck warning.

larvacea (approved):

PR #209, opened 2024-10-01
Add FIPS support for Rust binaries

arnaldo2792:
**Issue number:**
N / A

**Description of changes:**

This includes the changes required to build Rust binaries with FIPS support.

The first commit introduces the macros required to compile FIPS and non-FIPS binaries for both statically and dynamically compiled crates, as well as a new macro to compile binaries that depend on the `aws-sdk` . 

The `aws-lc-rs` crate's build script compiles `aws-lc` whenever the crate is build. The build script will pass the `CMAKE_TOOLCHAIN_FILE` environment variable to `cmake` in order to configure the cross compiler that should be used to compile the library. With this change, the `cargo_prep` macro generates and uses the `cmake` toolchain files for both static and dynamic builds. As part of this change, the `cross_cmake` macro was migrated to the new macros that generate the toolchain file.

All the FIPS binaries are created at `${HOME}/.cache/.fips/<type>` instead of having a `.fips` directory for every build type given that in the latter structure, the static builds would collide with each other and the FIPS build would fail with `No such file or directory errors` .

 The second commit in the series extends the `check-fips` script to detect whenever a Rust binary should include the extra FIPS copy. The script detects if a binary depends on `ring` but not in `aws-lc-rs`. This is to prevent relying on `ring` as the crypto backend for the rust programs included in Bottlerocket. Additionally, the binary detects when either the fips and non-fips binaries were compiled with the wrong `aws-lc-rs` flag, similar to the existing `go` check. 

**Testing done:**

- In a local change, I confirmed that `pluto` was compiled successfully after I migrated it to `aws-smithy-experimental`
- In a local change, I confirmed that the `apiclient` compiled successfully, with FIPS support
- In a local build, I confirmed the build failed to pass the `check-fips` check due to the missing `fips` binaries for a few packages
- In a local build, I confirmed that `docker-init` (the only package that depends on `cross_cmake`) compiled successfully

**Terms of contribution:**

### DIFF
--- a/macros/cargo
+++ b/macros/cargo
@@ -16,15 +16,45 @@
 %__cargo_cross_opts %{__cargo_common_opts} --target %{__cargo_target}
 %__cargo_cross_opts_static %{__cargo_common_opts} --target %{__cargo_target_static}
 %__cargo_env CARGO_TARGET_DIR="${HOME}/.cache" SKIP_README="true"
+%__cargo_env_fips CARGO_TARGET_DIR="${HOME}/.cache/.fips" SKIP_README="true"
+%__cargo_env_aws_sdk CARGO_TARGET_DIR="${HOME}/.cache/.aws-sdk" SKIP_README="true" RUSTFLAGS="%{__global_rustflags_aws_sdk}"
+%__cargo_env_fips_aws_sdk CARGO_TARGET_DIR="${HOME}/.cache/.fips/.aws-sdk" SKIP_README="true" RUSTFLAGS="%{__global_rustflags_aws_sdk}"
 %__cargo_env_static CARGO_TARGET_DIR="${HOME}/.cache/.static" SKIP_README="true"
+%__cargo_env_static_fips CARGO_TARGET_DIR="${HOME}/.cache/.fips/.static" SKIP_README="true"
 %__cargo_outdir "${HOME}/.cache/%{__cargo_target}/release"
+%__cargo_outdir_fips "${HOME}/.cache/.fips/%{__cargo_target}/release"
+%__cargo_outdir_aws_sdk "${HOME}/.cache/.aws-sdk/%{__cargo_target}/release"
+%__cargo_outdir_aws_sdk_fips "${HOME}/.cache/.fips/.aws-sdk/%{__cargo_target}/release"
 %__cargo_outdir_static "${HOME}/.cache/.static/%{__cargo_target_static}/release"
+%__cargo_outdir_static_fips "${HOME}/.cache/.fips/.static/%{__cargo_target_static}/release"
 %__cargo_cross_pkg_config PKG_CONFIG_PATH="%{_cross_pkgconfigdir}" PKG_CONFIG_ALLOW_CROSS=1
-%__cargo_cross_env %{__cargo_env} %{__cargo_cross_pkg_config} TARGET_CC="%{_cross_triple}-gnu-gcc"
-%__cargo_cross_env_static %{__cargo_env_static} %{__cargo_cross_pkg_config} TARGET_CC="%{_cross_triple}-musl-gcc"
+
+%__cargo_cross_env_base \
+  CMAKE_TOOLCHAIN_FILE="%{_cross_cmake_toolchain_conf}" ; \
+  export CMAKE_TOOLCHAIN_FILE ; \
+  BINDGEN_EXTRA_CLANG_ARGS="--target=%{_cross_target} --sysroot=%{_cross_rootdir}" ; \
+  export BINDGEN_EXTRA_CLANG_ARGS ; \
+  %{__cargo_cross_pkg_config}
+
+%__cargo_cross_env_static_base \
+  CMAKE_TOOLCHAIN_FILE="%{_cross_cmake_toolchain_conf_static}" ; \
+  export CMAKE_TOOLCHAIN_FILE ; \
+  BINDGEN_EXTRA_CLANG_ARGS="--target=%{_cross_triple}-musl --sysroot=/%{_cross_triple}-musl/sys-root" ; \
+  export BINDGEN_EXTRA_CLANG_ARGS ; \
+  %{__cargo_cross_pkg_config}
+
+%__cargo_cross_env %{__cargo_cross_env_base} %{__cargo_env} TARGET_CC="%{_cross_triple}-gnu-gcc"
+%__cargo_cross_env_aws_sdk %{__cargo_cross_env_base} %{__cargo_env_aws_sdk} TARGET_CC="%{_cross_triple}-gnu-gcc"
+%__cargo_cross_env_fips %{__cargo_cross_env_base} %{__cargo_env_fips} TARGET_CC="%{_cross_triple}-gnu-gcc"
+%__cargo_cross_env_fips_aws_sdk %{__cargo_cross_env_base} %{__cargo_env_fips_aws_sdk} TARGET_CC="%{_cross_triple}-gnu-gcc"
+%__cargo_cross_env_static %{__cargo_cross_env_static_base} %{__cargo_env_static} TARGET_CC="%{_cross_triple}-musl-gcc"
+%__cargo_cross_env_static_fips %{__cargo_cross_env_static_base} %{__cargo_env_static_fips} TARGET_CC="%{_cross_triple}-musl-gcc"
+
 %__cargo_incremental false
 
 %cargo_prep (\
+%write_cross_cmake_toolchain_conf \
+%write_cross_cmake_toolchain_conf_static \
 %{__mkdir} -p %{_builddir}/.cargo \
 cat > %{_builddir}/.cargo/config << EOF \
 [build]\
@@ -45,4 +75,8 @@ EOF\
 )
 
 %cargo_build %{__cargo_cross_env} %{__cargo} build %{__cargo_cross_opts} --release %{?cargo_args}
+%cargo_build_aws_sdk %{__cargo_cross_env_aws_sdk} %{__cargo} build %{__cargo_cross_opts} --release %{?cargo_args}
+%cargo_build_fips %{__cargo_cross_env_fips} %{__cargo} build %{__cargo_cross_opts} --release --features fips %{?cargo_args}
+%cargo_build_fips_aws_sdk %{__cargo_cross_env_fips_aws_sdk} %{__cargo} build %{__cargo_cross_opts} --release --features fips %{?cargo_args}
 %cargo_build_static %{__cargo_cross_env_static} %{__cargo} build %{__cargo_cross_opts_static} --release %{?cargo_args}
+%cargo_build_static_fips %{__cargo_cross_env_static_fips} %{__cargo} build %{__cargo_cross_opts_static} --release --features fips %{?cargo_args}
--- a/macros/check-fips
+++ b/macros/check-fips
@@ -45,13 +45,11 @@ uses_go_boring_crypto() {
   grep -Fq -m1 'goboringcrypto' <<<"${output}"
 }
 
-found=0
-not_found=0
-miscompiled=0
-for b in $(find "${BUILDROOT_BINDIR}" "${BUILDROOT_LIBEXECDIR}" -type f -executable) ; do
-  if is_go_bin "${b}" && uses_go_crypto_tls "${b}" ; then
-    f="${b/${BUILDROOT_BINDIR}/${BUILDROOT_FIPS_BINDIR}}"
-    f="${f/${BUILDROOT_LIBEXECDIR}/${BUILDROOT_FIPS_LIBEXECDIR}}"
+check_go_bin() {
+  local b f
+  b="${1:?}"
+  f="${2:?}"
+  if uses_go_crypto_tls "${b}" ; then
     if [ -s "${f}" ] ; then
       echo "${b} uses Go crypto/tls and FIPS build found in ${f}"
       (( found+=1 ))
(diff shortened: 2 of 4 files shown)


macros/cargo
```diff
@@ -16,11 +16,14 @@
 %__cargo_cross_opts %{__cargo_common_opts} --target %{__cargo_target}
 %__cargo_cross_opts_static %{__cargo_common_opts} --target %{__cargo_target_static}
 %__cargo_env CARGO_TARGET_DIR="${HOME}/.cache" SKIP_README="true"
+%__cargo_env_fips CARGO_TARGET_DIR="${HOME}/.cache/.fips" SKIP_README="true"
```
bcressey, line 19: It seems like you also need `__cargo_env_fips_static` (and related other macros) for cases like `apiclient` where both FIPS and static are in use.

macros/shared
```diff
@@ -152,6 +152,25 @@ CROSS_COMPILATION_CONF_EOF\
...
+set(CMAKE_C_COMPILER %{_cross_target}-gcc)\
+set(CMAKE_C_COMPILER_TARGET %_cross_target)\
+set(CMAKE_CXX_COMPILER %{_cross_target}-g++)\
+set(CMAKE_CXX_COMPILER_TARGET %_cross_target)\
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)\
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)\
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)\
+set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)\
+CROSS_CMAKE_TOOLCHAIN_EOF\
+%{nil}
```
bcressey, line 180: We only have one user of `%cross_cmake` - the `docker-init` package - so I'd prefer to go ahead and convert that macro to generate and use the toolchain file, the way that `%cross_meson` does.

macros/shared
```diff
@@ -225,6 +244,12 @@ CROSS_COMPILATION_CONF_EOF\
 %cross_scan_attribution \
   bottlerocket-license-scan --spdx-data /usr/libexec/tools/spdx-data --out-dir %{buildroot}%{_cross_attribution_vendor_dir}
 
+%set_cross_aws_lc_rs_env \
+  %write_cross_cmake_toolchain_conf \
+  export CMAKE_TOOLCHAIN_FILE="${PWD}/%{_cross_cmake_toolchain_conf}" \
+  export BINDGEN_EXTRA_CLANG_ARGS="--target=%{_cross_target} --sysroot=%{_cross_rootdir}" \
+  %{nil}
```
bcressey: I don't love this, since it's hard to imagine this combination being useful for anything other than `aws-lc-rs`, and these "shared" macros are meant to be reusable across specs.

Granted, you will need to build `aws-lc-rs` in multiple spec files, but you're still doing two unrelated things: configuring the `cmake` env and configuring the `bindgen` env.

It might be better to have two macros here:
```
%set_cross_cmake_env
%set_cross_bindgen_env
```

You could then have a third (in `macros/rust`) that calls them both, that could be used as a common preamble in packages that build Rust code, and we could add other global environment overrides there over time:
```
%set_cross_rust_env
```

macros/shared
```diff
@@ -225,6 +244,12 @@ CROSS_COMPILATION_CONF_EOF\
 %cross_scan_attribution \
   bottlerocket-license-scan --spdx-data /usr/libexec/tools/spdx-data --out-dir %{buildroot}%{_cross_attribution_vendor_dir}
 
+%set_cross_aws_lc_rs_env \
+  %write_cross_cmake_toolchain_conf \
+  export CMAKE_TOOLCHAIN_FILE="${PWD}/%{_cross_cmake_toolchain_conf}" \
+  export BINDGEN_EXTRA_CLANG_ARGS="--target=%{_cross_target} --sysroot=%{_cross_rootdir}" \
```
bcressey: Please follow the existing style here:
```suggestion
  CMAKE_TOOLCHAIN_FILE="${PWD}/%{_cross_cmake_toolchain_conf}" ; \
  export CMAKE_TOOLCHAIN_FILE ; \
  BINDGEN_EXTRA_CLANG_ARGS="--target=%{_cross_target} --sysroot=%{_cross_rootdir}" ; \
  export BINDGEN_EXTRA_CLANG_ARGS ; \
```

macros/cargo
```diff
@@ -44,5 +47,6 @@ rustflags = %{__global_rustflags_static_toml}\
 EOF\
 )
 
-%cargo_build %{__cargo_cross_env} %{__cargo} build %{__cargo_cross_opts} --release %{?cargo_args}
+%cargo_build %{?_cross_aws_lc_rs:%{set_cross_aws_lc_rs_env}} %{__cargo_cross_env}  %{__cargo} build %{__cargo_cross_opts} --release %{?cargo_args}
+%cargo_build_fips %{?_cross_aws_lc_rs:%{set_cross_aws_lc_rs_env}} %{__cargo_cross_env_fips}  %{__cargo} build %{__cargo_cross_opts} --release %{?cargo_args}
```
bcressey: Can we set something like `--features fips` in the `cargo` arguments to force the use of a FIPS feature, or error out if the target package doesn't understand it?

To me, `%cargo_build_fips` implies a stronger commitment to generating a binary that uses FIPS crypto, like `%cargo_build_static` does for static binaries. Just writing the artifacts to a different directory feels very weak, and it may confuse developers that expect it to do more than that.

macros/check-fips
```diff
@@ -68,6 +66,79 @@ for b in $(find "${BUILDROOT_BINDIR}" "${BUILDROOT_LIBEXECDIR}" -type f -executa
...
+
+check_rust_bin() {
+  local b f
+  b="${1:?}"
+  f="${2:?}"
+
+  if uses_ring "${b}" && ! uses_aws_lc_rs "${b}"; then
+    echo "${b} must use aws-lc-rs" >&2
+    (( miscompiled+=1 ))
+  fi
```
bcressey: When would binaries use both? That seems like an error also.

To cut down on the repeated `strings` calls on the same binaries, I would put this check at the end of the next one:
```
if uses_aws_lc_rs "${b}"; then
...
else
  # doesn't use aws_lc_rs
  if uses_ring "${b}"; then
    echo "${b} must use aws-lc-rs" >&2
    (( miscompiled+=1 ))
  fi
fi

File: macros/check-fips [hunk 7]
```diff
@@ -68,6 +66,79 @@ for b in $(find "${BUILDROOT_BINDIR}" "${BUILDROOT_LIBEXECDIR}" -type f -executa
...
+
+  if uses_aws_lc_rs "${b}"; then
+    if [ -s "${f}" ] ; then
+      echo "${b} uses aws-lc-rs and FIPS build found in ${f}"
+      (( found+=1 ))
+      if uses_aws_lc_fips_sys "${b}" ; then
+        echo "${b} is built with aws-lc-rs in FIPS mode rather than standard aws-lc-rs" >&2
+        (( miscompiled+=1 ))
+      fi
+      if ! aws_lc_fips_sys "${f}"; then
```
bcressey:
```suggestion
      if ! uses_aws_lc_fips_sys "${f}"; then
```

macros/cargo
arnaldo2792: Yeah, I wanted to keep that out from the initial review, I have it locally.

macros/cargo
arnaldo2792: We could, but AFAIU the `--features` flag can only be set once in `cargo` so we won't be able to override them if we ever have more `cargo` features. We could have a `_cargo_features` macro and for `cargo_build_fips` we could do `--features fips %{_cargo_features}`.

arnaldo2792: Resolved this in offline conversation, we will go with `--features fips` enforced in the macro.

arnaldo2792:
Forced push includes:
- `cargo_*` macros to compile `slow`, `fips`, `slow_fips`, and `static_fips` binaries
- New `__global_rustflags_slow` for slow rust builds (e.g. anything that depends on `aws-sdk`)
- Set of shareable macros to write a CMAKE toolchain file for dynamic and static builds
- Shareable macro to set the `bindgen` environment

arnaldo2792:
Forced push includes:
- In `cargo_static_fips`, I missed including `--features fips`
- Fix typo in `check-fips` script

macros/rust
```diff
@@ -11,6 +11,9 @@
 # Enable optimization, debuginfo, and link hardening.
 %__global_rustflags -Copt-level=3 -Cdebuginfo=2 -Ccodegen-units=1 -Cforce-frame-pointers=yes -Clink-arg=-Wl,-z,relro,-z,now
 
+# Disable optimization but keep debuginfo and link hardening
+%__global_rustflags_slow -Copt-level=3 -Cdebuginfo=2 -Cforce-frame-pointers=yes -Clink-arg=-Wl,-z,relro,-z,now
```
bcressey: The comment is wrong; we're not disabling optimization entirely, just one specific aspect of that.

I don't love "slow" as a suffix - it's actually slower to build with fewer codegen units, and in practice where we use these flags it's because we want the crate to build faster. This is the opposite of what the macro name implies.

Let's just use "aws_sdk" as the suffix. It's annoying to encode crate specifics into a macro but we end up doing that for "--features fips" so I can live with it to cut down on boilerplate elsewhere.

I would prefer to structure this as:
```
%__global_rustflags_base -Copt-level=3 -Cdebuginfo=2 -Cforce-frame-pointers=yes -Clink-arg=-Wl,-z,relro,-z,now
%__global_rustflags -Ccodegen-units=1 %__global_rustflags_base
%__global_rustflags_aws_sdk %__global_rustflags_base
```

macros/cargo
```diff
@@ -44,5 +56,9 @@ rustflags = %{__global_rustflags_static_toml}\
 EOF\
 )
 
-%cargo_build %{__cargo_cross_env} %{__cargo} build %{__cargo_cross_opts} --release %{?cargo_args}
-%cargo_build_static %{__cargo_cross_env_static} %{__cargo} build %{__cargo_cross_opts_static} --release %{?cargo_args}
+%cargo_build %set_cross_rust_env %{__cargo_cross_env}  %{__cargo} build %{__cargo_cross_opts} --release %{?cargo_args}
```
bcressey: I don't like using `%set_cross_rust_env` here; it means we'll be rewriting config files several times potentially, which is unexpected behavior for a compiler invocation.

We can run both the cmake toolchain macros as part of `%cargo_prep`, and set the `CMAKE_TOOLCHAIN_FILE` and `BINDGEN_EXTRA_CLANG_ARGS` variables to the right values in the `%_cargo_cross_env_*` macros.

macros/shared
```diff
@@ -225,6 +263,23 @@ CROSS_COMPILATION_CONF_EOF\
...
+%set_cross_cmake_env_static \
+  %write_cross_cmake_toolchain_conf_static \
+  CMAKE_TOOLCHAIN_FILE="${PWD}/%{_cross_cmake_toolchain_conf_static}" ; \
+  export CMAKE_TOOLCHAIN_FILE ; \
+  %{nil}
+
+%set_cross_bindgen_env \
+  BINDGEN_EXTRA_CLANG_ARGS="--target=%{_cross_target} --sysroot=%{_cross_rootdir}" ; \
+  export BINDGEN_EXTRA_CLANG_ARGS ; \
+  %{nil}
```
bcressey: The need for these goes away if you just move them into the respective `%_cargo` environment macros. But you're missing a `%set_cross_bindgen_env_static` if you decide to keep them.

macros/shared
```diff
@@ -152,6 +152,44 @@ CROSS_COMPILATION_CONF_EOF\
...
+set(CMAKE_SYSROOT /%{_cross_triple}-musl/sys-root)\
+set(CMAKE_C_COMPILER %{_cross_triple}-musl-gcc)\
+set(CMAKE_C_COMPILER_TARGET %{_cross_triple}-musl)\
+set(CMAKE_CXX_COMPILER %{_cross_triple}-musl-g++)\
+set(CMAKE_CXX_COMPILER_TARGET %{_cross_triple}-musl)\
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)\
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)\
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)\
+set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)\
+CROSS_CMAKE_TOOLCHAIN_EOF\
```
bcressey, line 205: I'd still like to see `%cross_cmake` updated to reflect this.

arnaldo2792: Yes, that was the last set of changes, I didn't want to play wack-a-mole with `docker-init` just yet.

arnaldo2792:
Forced push includes:
- Address feedback in last revision
- Change `cross_cmake` to use the `CMAKE_TOOLCHAIN_FILE` env variable and macros that generate the toolchain file
- Change the location of the `fips` generated artifacts for Rust binaries, because in the static builds of `apiclient` didn't play well together and some of the `fips` directories created by `cargo`(?) were deleted, resulting in `No file found` errors. After I moved them to `.fips/.static`, the problem went away.

macros/rust
```diff
@@ -8,8 +8,14 @@
...
-# Enable optimization, debuginfo, and link hardening.
-%__global_rustflags -Copt-level=3 -Cdebuginfo=2 -Ccodegen-units=1 -Cforce-frame-pointers=yes -Clink-arg=-Wl,-z,relro,-z,now
+# Enable some optimization, debuginfo, and link hardening.
+%__global_rustflags_base -Copt-level=3 -Cdebuginfo=2 -Cforce-frame-pointers=yes -Clink-arg=-Wl,-z,relro,-z,now
+
+# Optimize generated code by using one codegen unit
+%__global_rustflags -Ccodegen-units=1 %__global_rustflags_base
+
+# Pessimize AWS SDK binaries in favor of faster builds
+%__global_rustflags_aws_sdk %__global_rustflags_base
```
bcressey:
```suggestion
# Pessimize AWS SDK binaries in favor of faster builds
%__global_rustflags_aws_sdk -Cprefer-dynamic %__global_rustflags_base
```

macros/cargo
```diff
@@ -16,15 +16,45 @@
...
 %__cargo_cross_pkg_config PKG_CONFIG_PATH="%{_cross_pkgconfigdir}" PKG_CONFIG_ALLOW_CROSS=1
-%__cargo_cross_env %{__cargo_env} %{__cargo_cross_pkg_config} TARGET_CC="%{_cross_triple}-gnu-gcc"
-%__cargo_cross_env_static %{__cargo_env_static} %{__cargo_cross_pkg_config} TARGET_CC="%{_cross_triple}-musl-gcc"
+
+%__cargo_cross_env_base \
+  CMAKE_TOOLCHAIN_FILE="${PWD}/%{_cross_cmake_toolchain_conf}" ; \
+  export CMAKE_TOOLCHAIN_FILE ; \
+  BINDGEN_EXTRA_CLANG_ARGS="--target=%{_cross_target} --sysroot=%{_cross_rootdir}" ; \
+  export BINDGEN_EXTRA_CLANG_ARGS ; \
+  %{__cargo_cross_pkg_config}
```
bcressey: Are you confident the `"${PWD}"` here is doing the right thing? It seems like it would depend on when the macro was evaluated.

It might be better to explicitly write the cmake toolchain configs to `%{_builddir}` so we can use that as a canonical directory, like we do for `%{_builddir}/.cargo/config`.

arnaldo2792:
Forced push includes:
- Change `_cross_cmake_toolchain_conf*` path to be in `%{_builddir}`
- Add extra flag to `%__global_rustflags_aws_sdk`

bcressey (approved):

arnaldo2792:
Forced push includes:
- Fix path for `CMAKE_TOOLCHAIN_FILE` in `cross_cmake` macro
- Remove unnecessary lines in `check-fips` script

bcressey (approved):

ginglis13 (approved):

cbgbt (approved):

PR #210, opened 2024-10-02
macros, build: override find-debuginfo location

bcressey:
**Issue number:**
N/A

**Description of changes:**
This fixes a regression where `find-debuginfo` now prefers `objcopy` in `/usr/bin` rather than the target-specific symlink placed ahead of it in $PATH. That happens because `find-debuginfo` now adds its own parent directory to $PATH in front of everything else.

Work around this by adding a `find-debuginfo` symlink to the debug helper directory, and override the macro so that `find-debuginfo` is called from that directory.

Also add a `readelf` symlink, since `find-debuginfo` calls it too.

**Testing done:**
Before, the output logs would show errors like this when building on `x86_64` for `aarch64`:
```
  #14 133.7 find-debuginfo: starting
  #14 133.8 Extracting debug info from 9 files
  #14 134.9 objcopy: Unable to recognise the format of the input file `/home/builder/rpmbuild/BUILDROOT/bottlerocket-settings-plugins-0.0-0.1727497932.29b947  0bc.br1.x86_64/aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/settings-plugins/aws-dev/libsettings.so'
```

After, the errors go away since the target-specific `objcopy` is being used.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1183,15 +1183,19 @@ RUN \
 
 # =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 #
-# Create symlinks that can be added to $PATH to override programs invoked by
-# find-debuginfo.sh, which does not expect to add a prefix.
+# find-debuginfo adds the directory where it's located as the first entry in $PATH.
+# Set up a directory for each architecture containing the unprefixed versions of
+# the binutils programs, along with a link to find-debuginfo, so that it will work
+# as expected when extracting and stripping debuginfo from binaries built for the
+# target architecture.
 FROM sdk as sdk-find-debuginfo-symlinks
 RUN \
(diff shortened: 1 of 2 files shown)


bcressey:
This is what feels like the millionth time I've had to fix debuginfo extraction in our cross-compilation setup.

That said, we will now run `find-debuginfo` from a specially-prepared, target-specific directory with controlled symlinks that point to the correct binutils programs to use. So this change should be just about as self-contained and focused on the precise use case - debuginfo extraction - as is possible, without the negative side effects of cluttering `$PATH` during the build.

See https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/174 for a related fix that this helped to catch, after the cleanup in https://github.com/bottlerocket-os/twoliter/pull/383.

arnaldo2792 (approved):

yeazelm (approved):

PR #211, opened 2024-10-04
license-scan: prepare to publish to crates.io

cbgbt:
**Description of changes:**
* Add copy of project licenses to crate
* Add additional useful crate metadata to Cargo.toml

My intent is to follow this PR up with integration of [release-plz](https://github.com/MarcoIeni/release-plz) for automated crate releases.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/test-license-scan.yml
+++ b/.github/workflows/test-license-scan.yml
@@ -1,0 +1,19 @@
+name: Test bottlerocket-license-scan
+on:
+    pull_request:
+        branches:
+            - "*"
+        paths:
+            - "license-scan/**"
+            - ".github/workflows/test-license-scan.yml"
+jobs:
+    build:
+        runs-on:
+            group: bottlerocket
+            labels: bottlerocket_ubuntu-latest_8-core
+        steps:
+            - uses: actions/checkout@v3
+            - run: rustup update stable
+            - run: cargo install cargo-deny --locked
+            - run: make build
+              working-directory: ./license-scan
--- a/license-scan/Cargo.toml
+++ b/license-scan/Cargo.toml
@@ -1,10 +1,13 @@
 [package]
 name = "bottlerocket-license-scan"
 version = "0.1.0"
+description = "Collects and identifies licenses of Rust and Golang dependencies"
 license = "Apache-2.0 OR MIT"
 authors = ["iliana destroyer of worlds <iweller@amazon.com>"]
+repository = "https://github.com/bottlerocket-os/bottlerocket-sdk"
 edition = "2018"
-publish = false
+keywords = ["license", "spdx", "ci"]
+readme = "README.md"
 
 [dependencies]
 anyhow = "1"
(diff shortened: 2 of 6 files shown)


license-scan/Cargo.toml
```diff
@@ -1,10 +1,13 @@
 [package]
 name = "bottlerocket-license-scan"
 version = "0.1.0"
+description = "Collects and identifies licenses of Rust and Golang dependencies"
 license = "Apache-2.0 OR MIT"
 authors = ["iliana destroyer of worlds <iweller@amazon.com>"]
+repository = "https://github.com/bottlerocket-os/bottlerocket-sdk"
 edition = "2018"
```
jpculp, line 8:
```suggestion
edition = "2021"
```
Should we bump our edition and rust dependencies before we publish?

cbgbt: Yes, a dependency bump is definitely in order.

We don't need to increment the edition as a blocker here though, Rust allows editions to peacefully coexist.

cbgbt:
^ bump dependencies per @jpculp.

For whatever reason my new GitHub action won't run for this PR, but I'm hopeful that it's because it will run hereafter.

arnaldo2792 (approved):

.github/workflows/test-license-scan.yml
```diff
@@ -0,0 +1,16 @@
+name: Test bottlerocket-license-scan
+on:
+    pull_request:
+        branches:
+            - "*"
+        paths:
+            - "license-scan/**"
+jobs:
+    build:
+        runs-on: ubuntu-latest
```
bcressey: I'm not sure we can use `ubuntu-latest` as a runner; it might need to be shared from the org level.

I noticed that the only core kit workflow to use this runner - [golangci-lint.yaml](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/.github/workflows/golangci-lint.yaml) doesn't seem to have triggered for a [recent PR](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/146) - so this might be a new change.

Try changing it to this and see if the workflow runs:
```suggestion
        runs-on:
            group: bottlerocket
            labels: bottlerocket_ubuntu-latest_8-core
```

bcressey (approved):

cbgbt:
I've tried all manner of reconfiguring this action and peering into the repository configurations and cannot for the life of me get the action to run in this PR.

@bcressey another possible culprit in the core-kit PR is that we possibly need double-wildcards [here](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/403ae6ececa6d577f17b980532d59d1b4e48d073/.github/workflows/golangci-lint.yaml#L8).

In any case, I'd appreciate another review. I'll open a test PR after this one is merged to see if it triggers the workflow.

bcressey (approved):

bcressey:
> In any case, I'd appreciate another review. I'll open a test PR after this one is merged to see if it triggers the workflow.

Sounds good, let's see what happens!

PR #222, opened 2024-11-14
include glibc in the final SDK

### LINKED ISSUE #221: include glibc headers and libraries in SDK
**Image I'm using:**
v0.47.0

**Issue or Feature Request:**
I'd like to include the glibc libraries and headers in the SDK, as we already do for musl.

It's still desirable to ship glibc as packaged by the core kit, for ease of patching and because of some complexities involved in creating the final build (such as dealing with locales).

However, providing a copy of glibc in the SDK's sys-roots will make it easier to move various packages out of the core kit. This is especially true for Go and Rust programs, which rarely need anything beyond the C library to build. It also helps with packages like `libkcapi`, which would ideally go along [when the kernels move to a separate kit](https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/193).

This relies a lot on glibc's backwards compatibility, and on making sure the SDK's version of glibc never gets ahead of the core kit's.

**Alternatives:**
Twoliter supports kits that depend on other kits, so we could just make everything depend on the core kit. This wouldn't help for `libkcapi`, and it would be overkill for kits that only needed the glibc dependency.

We could also have a libc kit. In practice, that would take us from the current world where all kits have to be updated whenever the SDK changes, to a world where all kits have to take a combined SDK and libc kit update. Keeping the two together is appealing in the absence of fully automated updates everywhere.

Also, there's overhead and inefficiency that arises from making changes that touch multiple kits. I'd like to stick with larger building blocks for now: kits for "all kernels" and "all of userspace" are OK, but kits for one library are not.

bcressey:
**Issue number:**
Fixes #221

**Description of changes:**
Include glibc headers and libraries in the final SDK.

**Testing done:**
Built the core kit with these changes. This required a small change to the kernel configs, since now the kernel detects a working environment for building userspace programs.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1217,10 +1217,11 @@ COPY --from=sdk-plus-lz4 /usr/share/licenses/lz4 /usr/share/licenses/lz4
 # a format that's convenient for extracting later.
 COPY --from=toolchain-archive /toolchain /toolchain
 
-# "sdk-libc-musl" has the musl C library and headers. We omit "sdk-libc-gnu"
-# because we expect to build glibc again for the target OS, while we will use
-# the musl artifacts directly to generate static binaries such as migrations.
+# "sdk-libc-musl" has the MUSL C library and headers.
 COPY --from=sdk-libc-musl / /
+
+# "sdk-libc-gnu" has the GNU C library and headers.
+COPY --from=sdk-libc-gnu / /
 
 # "sdk-rust" has our Rust toolchain with the required targets.
 COPY --chown=0:0 --from=sdk-rust /usr/libexec/rust/ /usr/libexec/rust/


jmt-lab (approved):

yeazelm (approved):

Dockerfile
```diff
@@ -1217,11 +1217,12 @@ COPY --from=sdk-plus-lz4 /usr/share/licenses/lz4 /usr/share/licenses/lz4
...
 COPY --from=toolchain-archive /toolchain /toolchain
 
-# "sdk-libc-musl" has the musl C library and headers. We omit "sdk-libc-gnu"
-# because we expect to build glibc again for the target OS, while we will use
-# the musl artifacts directly to generate static binaries such as migrations.
+# "sdk-libc-musl" has the MUSL C library and headers.
 COPY --from=sdk-libc-musl / /
 
+# "sdk-libc-gnu" has the GNU C library and headers.
+COPY --from=sdk-libc-gnu / /
```
arnaldo2792, line 1224: > This relies a lot on glibc's backwards compatibility, and on making sure the SDK's version of glibc never gets ahead of the core kit's.

Is there going to be a check similar to the check done for `binutils` to prevent the `glibc` copy in the core kit from getting out of sync with the `glibc` version in the SDK?

arnaldo2792 (approved):

Dockerfile
bcressey: > Is there going to be a check similar to the check done for binutils to prevent the glibc copy in the core kit from getting out of sync with the glibc version in the SDK?

I expect it to be OK for the SDK copy of glibc to be older; it just can't be newer.

PR #224, opened 2024-11-20
Update to binutils 2.41

larvacea:
Update buildroot to 2023.08.4 and binutils to 2.41.

**Description of changes:**

Update buildroot to 2023.08.4 in order to update binutils to 2.41. Once this change is live, core kit will need an update to its packages/binutils to upgrade that to 2.41 (which I will supply in a pull request on that repo).

**Testing done:**

Built and published the SDK to private ECR. Built core kit on both aarch64 and x86_64 hosts, for both target architectures.

**Terms of contribution:**


patches/buildroot/0001-disable-shared-for-host-builds-of-gmp-isl-mpc-mpfr.patch
```diff
@@ -1,7 +1,7 @@
-From 9e6d1c176a007d6751e3acd02f96becee48ef361 Mon Sep 17 00:00:00 2001
+From b90b648d30fe4b6e6554a04b9d3f7d93aebdf6c7 Mon Sep 17 00:00:00 2001
 From: Ben Cressey <bcressey@amazon.com>
 Date: Sun, 6 Aug 2017 23:15:02 +0000
-Subject: [PATCH] disable shared for host builds of gmp, isl, mpc, mpfr
+Subject: [PATCH 1/6] disable shared for host builds of gmp, isl, mpc, mpfr
```
arnaldo2792: You should remove the patch numbers with `--no-numbered`

patches/buildroot/0005-Bring-back-sysroot-fix-for-binutils.patch
```diff
@@ -0,0 +1,195 @@
+From 9afc3195d55b209fbf6668ed159198a930b9163b Mon Sep 17 00:00:00 2001
+From: Martin Harriman <larvacea@mac.com>
+Date: Mon, 18 Nov 2024 23:00:05 +0000
+Subject: [PATCH 5/6] Bring back sysroot fix for binutils
```
arnaldo2792: Did they mention why they removed these patches?

larvacea: I do not know. This patch-within-a-patch is present in the current SDK. It had to change to this form because it applies to a versioned subdirectory in buildroot. The current patch applied to the 2.38 subdirectory, no longer present as of buildroot 2023.08.4, and needs now to apply to the versioned subdirectories in buildroot 2023.08.4.

patches/buildroot/0001-disable-shared-for-host-builds-of-gmp-isl-mpc-mpfr.patch
larvacea: Done

Dockerfile
```diff
@@ -83,8 +83,8 @@ RUN \
   git config --global user.email "builder@localhost"
 
 ARG UPSTREAM_SOURCE_FALLBACK
-ENV BRVER="2022.11.1"
-ENV KVER="5.10.162"
+ENV BRVER="2023.08.4"
+ENV KVER="5.10.202"
```
bcressey: We can go all the way to `2024.02.8` while keeping GCC 11, and should as part of this update.

patches/buildroot/0002-add-TOOLS_DIR-and-SYSROOT_DIR-to-control-output.patch
```diff
@@ -1,6 +1,6 @@
-From c137f14fade2ae3c23ab389b2e3e99f8bb5a55ab Mon Sep 17 00:00:00 2001
-From: Ben Cressey <bcressey@amazon.com>
-Date: Sun, 6 Aug 2017 23:35:06 +0000
+From 042b3f15465d1d93e0b99e0d659c7cb55eef3b15 Mon Sep 17 00:00:00 2001
+From: Martin Harriman <larvacea@mac.com>
```
bcressey: Please keep the author line intact; you can add notes below my sign-off and then add your own sign-off to capture your contribution.

E.g.
```
Signed-off-by: Ben Cressey <bcressey@amazon.com>
[larvacea:
  - rebase to x.y.z
  - fixed context
  - reworked blah
]
Signed-off-by: Martin Harriman <larvacea@mac.com>
```

patches/buildroot/0004-build-gcc-with-TOOLS_DIR-and-SYSROOT_DIR.patch
```diff
@@ -190,18 +190,18 @@ index ebd8e8f017..ec9fcd13c1 100644
...
-index ec6f3b550a..6d4f6e0b20 100644
+index 834da249..bfb7de9a 100644
 --- a/package/gcc/gcc.mk
 +++ b/package/gcc/gcc.mk
-@@ -69,7 +69,6 @@ HOST_GCC_COMMON_DEPENDENCIES = \
+@@ -68,7 +68,6 @@ HOST_GCC_COMMON_DEPENDENCIES = \
  
  HOST_GCC_COMMON_CONF_OPTS = \
  	--target=$(GNU_TARGET_NAME) \
 -	--with-sysroot=$(STAGING_DIR) \
```
bcressey, line 199: This should stay; not sure why it went away.

patches/buildroot/0005-Bring-back-sysroot-fix-for-binutils.patch
```diff
@@ -0,0 +1,195 @@
+From 9afc3195d55b209fbf6668ed159198a930b9163b Mon Sep 17 00:00:00 2001
```
bcressey: Generally I'd recommend only adding the patch for the binutils version we're using; otherwise it'll get tripped up on rebase if the older binutils versions are dropped.

But that's a minor hiccup and here I think the value is that we could revert to an older binutils more easily if the 2.41 update causes problems.

patches/buildroot/0006-Add-patch-series-for-stack-protection.patch
```diff
@@ -2686,7 +2686,7 @@ index 0000000000..d5ab90340d
 +
 diff --git a/package/gcc/11.3.0/9019-aarch64-Make-stack-smash-canary-protect-saved-registers.patch b/package/gcc/11.3.0/9019-aarch64-Make-stack-smash-canary-protect-saved-registers.patch
 new file mode 100644
-index 0000000000..33866e8afd
+index 00000000..33866e8a
 --- /dev/null
 +++ b/package/gcc/11.3.0/9019-aarch64-Make-stack-smash-canary-protect-saved-registers.patch
```
bcressey: This patch series ends up being lost because GCC was updated from 11.3.0 to 11.4.0, so the patch files aren't picked up during the build.

bcressey:
I've pushed my rebased patches for Buildroot 2024.02.8 [here](https://github.com/bcressey/bottlerocket-sdk/tree/buildroot-bump).

Relative to that work, I'm fine with your move to binutils 2.41 (as long as it works downstream), since that provides more clarity on which CVEs have been patched.

I'd like to bring over the new GCC patch to fix `--disable-fixincludes`, since I need that disabled for a downstream update to glibc 2.40.

Dockerfile
larvacea: This will also bring in glibc 2.38 and add a patch to aws-lc to fix compilation errors in two tests (the patch backports changes from the tests in main-branch aws-lc).

larvacea:
Updated.
Summary of package changes relative to develop:
* buildroot 2024.02.8
* kernel 5.10.228
* glibc 2.38
* binutils 2.41
* gawk 5.3.0
* gcc 11.4.0
* gmp 6.3.0
* isl 0.26
* musl 1.2.4
* mpfr 4.1.1

Patches:
* aws-lc: patch two tests to fix compilation errors (backported from aws-lc main branch).
* buildroot: updated to include @bcressey's latest patch and ensure that the patches apply to the correct versioned subdirectories.

Testing:
* The SDK builds on both host architectures and uploads a valid manifest once both are uploaded to a private ECR.
* Core kit builds on both host architectures for both target architectures (using this SDK, and with a change to the binutils package to match the SDK binutils version).
* Bob builds on both host architectures for both target architectures.
* AMIs boot on both target architectures.

Dockerfile
```diff
@@ -184,8 +184,8 @@ WORKDIR /home/builder
 COPY ./hashes/glibc ./hashes
 COPY ./helpers/glibc/* ./
 
-ENV GLIBCVER="2.37"
-ENV KVER="5.10.162"
+ENV GLIBCVER="2.38"
```
bcressey, line 187: We should update the musl build to 1.2.4 to match the newer Buildroot version.

bcressey (approved):

larvacea:
Updated the SDK musl to match buildroot's version.

Dockerfile
```diff
@@ -1217,11 +1217,12 @@ COPY --from=sdk-plus-lz4 /usr/share/licenses/lz4 /usr/share/licenses/lz4
...
 COPY --from=toolchain-archive /toolchain /toolchain
 
-# "sdk-libc-musl" has the musl C library and headers. We omit "sdk-libc-gnu"
-# because we expect to build glibc again for the target OS, while we will use
-# the musl artifacts directly to generate static binaries such as migrations.
+# "sdk-libc-musl" has the MUSL C library and headers.
 COPY --from=sdk-libc-musl / /
 
+# "sdk-libc-gnu" has the GNU C library and headers.
+COPY --from=sdk-libc-gnu / /
```
bcressey: Why is this showing up as a change here? It was merged in https://github.com/bottlerocket-os/bottlerocket-sdk/commit/8ef6f479f.

larvacea: My best guess is that either I (more likely) or git (less likely) got bewildered in the rebase process. I have rebased on develop to pick up the base image change, and checked that this is not showing up as a change in the commit.

jpculp (approved):

bcressey (approved):

bcressey:
I am seeing this build failure in core kit with these changes:
```
  #14 142.1 + gofips build -ldflags '-compressdwarf=false -linkmode=external -extldflags '\''-Wl,-z,relro -Wl,-z,now -Wl,-Bsymbolic-functions -static-pie'\''' -o fips-static/amazon
-ssm-agent ./core/agent.go ./core/agent_unix.go ./core/agent_parser.go
  #14 149.8 # crypto/internal/boring
  #14 149.8 cgo-gcc-prolog: In function ‘_cgo_15785f09c110_Cfunc__goboringcrypto_BORINGSSL_bcm_power_on_self_test’:
  #14 149.8 cgo-gcc-prolog:141:49: warning: unused variable ‘_cgo_a’ [-Wunused-variable]                                                                                              #14 160.7 # command-line-arguments                                                                                                                                                  #14 160.7 /usr/libexec/go-1.22/pkg/tool/linux_amd64/link: running x86_64-bottlerocket-linux-musl-gcc failed: exit status 1                                                          #14 160.7 /usr/lib/gcc/x86_64-bottlerocket-linux-musl/11/../../../../x86_64-bottlerocket-linux-musl/bin/ld: /tmp/go-link-1440672810/000032.o: in function `BIO_get_fd':             #14 160.7 (.text+0xe84f): undefined reference to `fopen64'                                                                                                                          #14 160.7 /usr/lib/gcc/x86_64-bottlerocket-linux-musl/11/../../../../x86_64-bottlerocket-linux-musl/bin/ld: /tmp/go-link-1440672810/000032.o: in function `BIO_new_file':
  #14 160.7 (.text+0xea44): undefined reference to `fopen64'
  #14 160.7 /usr/lib/gcc/x86_64-bottlerocket-linux-musl/11/../../../../x86_64-bottlerocket-linux-musl/bin/ld: /tmp/go-link-1440672810/000032.o: in function `bn_mul_mont':
  #14 160.7 (.text+0x25d3e8): undefined reference to `__isoc23_sscanf'
  #14 160.7 collect2: error: ld returned 1 exit status
```

My guess is that it's one of these two changes, which I haven't previously tested:
* the update to Fedora 40, which has a glibc with those symbols
* the update to binutils, if it's incorrectly linking with the host glibc because of some issue with the patch

Ultimately the root cause here is not either of those, it's that we take a glibc-linked AWS-LC DSO and try to use it in a static Go binary with musl. But it might be faster to revert one of the other changes.

bcressey:
> My guess is that it's one of these two changes, which I haven't previously tested ...

Update: it's not either of those two changes 😢

bcressey:
Opened https://github.com/bottlerocket-os/bottlerocket-sdk/issues/225 to track the issue. It's not a blocker for merging this PR, but it's a blocker for the next SDK release.

PR #241, opened 2025-02-03
macros: add compat symlink for workspace sources

bcressey:
**Issue number:**
Related: #239

**Description of changes:**
This is a better version of the initial fix in #239 - rather than expecting downstream packages to switch to the new macro, create a symlink to the old location for workspace sources in the per-package `%{builddir}`.

The old location is widely used by the core kit and downstream variant builds, and breaking it would create unnecessary friction for adopting an updated SDK.

**Testing done:**
Built the core kit using an SDK with this change, plus the switch to a Fedora 41 base image with RPM 4.20.

**Terms of contribution:**

### DIFF
--- a/macros/shared
+++ b/macros/shared
@@ -325,6 +325,14 @@ CROSS_CMAKE_TOOLCHAIN_EOF\
   %{?cross_check_fips:%{__cross_check_fips_cmd}} \
   %cross_generate_attribution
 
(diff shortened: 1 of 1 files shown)


jpculp (approved):

cbgbt (approved):
Good idea!

PR #244, opened 2025-02-13
build: update base image to Fedora 41

### LINKED ISSUE #238: update base image to Fedora 41
**Image I'm using:**
v0.50.0

**Issue or Feature Request:**
I'd like to update the base image from Fedora 39 to 41, which includes various improvements like RPM 4.20.

bcressey:
**Issue number:**
Fixes: #238

**Description of changes:**
The largest breaking change from the Fedora 41 update is covered by #241, which is safe to merge independently from this, and therefore doesn't need to be reverted if this is reverted.

That leaves a bunch of minor changes:
* Deal with the behavior change in `rpm2cpio` from RPM 4.20: files are now prefixed with "./", and when extracting a subset of files, this prefix must be provided or no match will be found.
* Undefine %_auto_set_build_flags in the default macros, since it's no longer trivial to do so downstream, and they clash with the required build flags for cross-compiling.
* efitools is not used downstream and is no longer available in Fedora, so drop it.
* The OpenSSL ENGINE API headers are no longer installed by default, so add them back. They can be dropped again when we update the AWS SDK for C++ with a new-enough version of s2n, which is not yet available.
* The gold linker is no longer installed by default, so add a Go patch so that it's not preferred on aarch64.

**Testing done:**
Built kits and variants, albeit with some minor changes required downstream.

**Terms of contribution:**

### DIFF
--- a/patches/go-1.22/0003-Default-to-ld.bfd-on-ARM64.patch
+++ b/patches/go-1.22/0003-Default-to-ld.bfd-on-ARM64.patch
@@ -1,0 +1,46 @@
+From 46ec67413008607e2150e3395668e54e538c5b6b Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Alejandro=20S=C3=A1ez?= <asm@redhat.com>
+Date: Wed, 19 Jun 2024 10:18:58 +0200
+Subject: [PATCH] Default to ld.bfd on ARM64
+
+---
+ src/cmd/link/internal/ld/lib.go | 20 +++++++-------------
+ 1 file changed, 7 insertions(+), 13 deletions(-)
+
+diff --git a/src/cmd/link/internal/ld/lib.go b/src/cmd/link/internal/ld/lib.go
+index eab74dc328..b401f58727 100644
+--- a/src/cmd/link/internal/ld/lib.go
++++ b/src/cmd/link/internal/ld/lib.go
+@@ -1620,22 +1620,16 @@ func (ctxt *Link) hostlink() {
+ 		}
+ 
+ 		if ctxt.Arch.InFamily(sys.ARM64) && buildcfg.GOOS == "linux" {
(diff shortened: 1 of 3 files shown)


patches/go-1.22/0003-Default-to-ld.bfd-on-ARM64.patch
```diff
@@ -0,0 +1,46 @@
...
+-			// required. Switch to gold.
+-			// https://sourceware.org/bugzilla/show_bug.cgi?id=19962
+-			// https://go.dev/issue/22040
+-			altLinker = "gold"
+-
+-			// If gold is not installed, gcc will silently switch
+-			// back to ld.bfd. So we parse the version information
+-			// and provide a useful error if gold is missing.
++			// Use ld.bfd as the default linker
++			altLinker = "bfd"
```
cbgbt, line 29: Given the context in the linked issues, I assume upstream will probably also converge back to `bfd`. Seems like the reason for the switch to `gold` in the first place is resolved upstream, and `gold` is on a deprecation path.

cbgbt (approved):
>  Built kits and variants, albeit with some minor changes required downstream.

Is this referring to https://github.com/bottlerocket-os/twoliter/pull/457? Or are there other potential backwards incompatibilities here?

bcressey:
> Is this referring to https://github.com/bottlerocket-os/twoliter/pull/457? Or are there other potential backwards incompatibilities here?

I've related all the downstream pull requests to this one. Besides Twoliter, the kernel kit took the brunt of it - lots of `rpm2cpio` usage plus one of the kernel configs was using SHA-1 which is now blocked by the default security policy.

jpculp (approved):

bcressey:
⬆️ force push to rebase

sam-berning (approved):

PR #266, opened 2025-04-10
aws-lc: bump from 2.0.17 to 3.0.0

ytsssun:
**Description of changes:**
Bump `aws-lc` 2.0.17 to `3.0.0`

**Testing done:**
Tested building custom FIPS and non-FIPS AMI with this SDK change. 
I have a setup where I use an nginx pod to enforce cipher suite that is not supported by FIPS
```
# Not supported by FIPS
ssl_ciphers 'ECDHE-RSA-CHACHA20-POLY1305';
```

And use ctr images pull to verify the tls behavior.

For FIPS variant, the tls handshake failed as expected:
```
bash-5.1# ctr images pull localhost:8043/test:latest
INFO[0000] host will try HTTPS first since it is configured for HTTP with a TLS configuration, consider changing host to HTTPS or removing unused TLS configuration  host="localhost:8043"
INFO[0000] trying next host                              error="failed to do request: Head \"https://localhost:8043/v2/test/manifests/latest\": remote error: tls: handshake failure" host="localhost:8043"
ctr: failed to resolve reference "localhost:8043/test:latest": failed to do request: Head "https://localhost:8043/v2/test/manifests/latest": remote error: tls: handshake failure
```

For non-FIPS variant, the tls handshake succeeded and the command is returning the normal image not found response:
```
ctr images pull localhost:8043/test:latest
INFO[0000] host will try HTTPS first since it is configured for HTTP with a TLS configuration, consider changing host to HTTPS or removing unused TLS configuration  host="localhost:8043"
INFO[0000] trying next host - response was http.StatusNotFound  host="localhost:8043"
ctr: failed to resolve reference "localhost:8043/test:latest": localhost:8043/test:latest: not found
```

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #270, opened 2025-05-13
serialize aws-lc builds inside a Go version

bcressey:
**Issue number:**
Related: #266

**Description of changes:**
The delocate phase of the aws-lc build can use up to 25 GiB of RAM, and running eight of them in parallel can lead to OOM even for build hosts with lots of memory.

Force each stage within a Go version to depend on the stage for the previous build, so that at most two versions will build concurrently.

**Testing done:**
Built locally.

**Terms of contribution:**


ytsssun (approved):

jpculp (approved):
I tried building https://github.com/bottlerocket-os/bottlerocket-sdk/pull/266 along with this change and did not hit OOM! 🚀

PR #275, opened 2025-05-22
update Rust to 1.87.0

### LINKED ISSUE #272: update to Rust 1.87.0
[Rust 1.87.0 is out!](https://blog.rust-lang.org/2025/05/15/Rust-1.87.0/)

bcressey:
**Issue number:**
Fixes #272

**Description of changes:**
Update Rust to 1.87.0.

Set the panic strategy to "abort" for all custom targets. This preserves unwinding for developer builds outside of the SDK, where the custom targets are not available, while reducing the disk usage and runtime overhead for production builds.

**Testing done:**
Built kits and smoke-tested variants for both architectures.

**Terms of contribution:**


jmt-lab (approved):

Dockerfile
```diff
@@ -285,7 +285,7 @@ RUN \
 
 ARG HOST_ARCH
 ENV VENDOR="bottlerocket"
-ENV RUSTVER="1.85.1"
+ENV RUSTVER="1.86.0"
```
arnaldo2792: Do you need to keep the commits that bump 1.85.0 -> 1.85.1 -> 1.86.0 -> 1.87.0?

bcressey: I kept them to show the progression, and in case we needed to revert 1.87.0 at the last minute without necessarily backing out the other changes. 1.85.1 is useful as it lets us drop the `cc-rs` patch.

bcressey:
Marking as ready for review now that the downstream clippy fixes have been merged.

KCSesh (approved):

arnaldo2792 (approved):

PR #276, opened 2025-05-27
update compiler and linker flags

### LINKED ISSUE #265: sync compiler flags with recent recommendations
With GCC 13, we can enable some additional hardening flags like these:
https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html

bcressey:
**Issue number:**
Fixes #265

**Description of changes:**
Add additional hardening flags, as well as architecture-specific tuning flags. These are loosely aligned with Fedora's defaults, with the major exception being that the aarch64 flags assume the "crc" and "crypto" extensions. These are supported by all generations of Graviton instance types in EC2, but not necessarily for other aarch64 CPUs, where the extensions are optional.

**Testing done:**
Built kits and smoke-tested variants for both architectures.

Confirmed that older EC2 instance types - `c3.2xlarge`, `a1.xlarge` - still worked with the new compiler flags.

**Terms of contribution:**


macros/shared
```diff
@@ -7,14 +7,22 @@
...
   -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer \
+  -ftrivial-auto-var-init=zero \
   %{nil}}
+
 %_cross_cxxflags %_cross_cflags
-%_cross_ldflags -Wl,-z,relro -Wl,-z,now -Wl,-Bsymbolic-functions -Wl,-z,pack-relative-relocs
+
+%_cross_ldflags %{shrink: \
+  -Wl,-z,relro -Wl,-z,now \
+  -Wl,-z,noexecstack -Wl,--as-needed  \
```
arnaldo2792, line 25: With `--as-needed` did you see any reduction in the size of the Go binaries?

macros/shared
```diff
@@ -14,6 +14,7 @@
   -fexceptions -fstack-clash-protection -fno-semantic-interposition \
   -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer \
   -ftrivial-auto-var-init=zero \
+  -flto=auto -ffat-lto-objects \
```
arnaldo2792, line 17: Same question elsewhere, did you see any gains in the binary size of the Rust programs?

arnaldo2792 (approved):

macros/shared
```diff
@@ -7,21 +7,38 @@
...
-%_cross_c_link_args '-Wl,-z,relro', '-Wl,-z,now', '-Wl,-Bsymbolic-functions', '-Wl,-z,pack-relative-relocs'
+
+%_cross_ldflags %{shrink: \
+  -Wl,-z,relro -Wl,-z,now \
+  -Wl,-z,noexecstack -Wl,--as-needed  \
+  -Wl,-Bsymbolic-functions -Wl,-z,pack-relative-relocs \
+  %{_cross_arch_ldflags} \
+  %{nil}}
+
+%_cross_quoted_list() %{lua:
```
yeazelm, line 30: This is a nice improvement for maintenance!

yeazelm (approved):

macros/shared
bcressey: Go binaries are essentially the same size as they were before the change.

macros/shared
bcressey: Rust programs have been compiled with LTO for years now, and these flags aren't (and shouldn't be) passed through to C builds.

There was quite a significant reduction in Rust binary size though. I expect that's attributable to the upgrade to Rust 1.87.0 rather than any change here. (edit: and also the switch to the "abort" panic strategy)

For example, `pluto` went from 16.12 MiB to 12.04 MiB on aarch64, and from 17.93 MiB to 14.18 MiB on x86_64.

PR #283, opened 2025-06-26
Add sbomtool to SDK

rpkelly:
**Description of changes:**
Created a tool to generate SBOMs as well as a macro to run the tool on package creation.

**Testing done:**
* Ran unit tests
* Built SDK
* Updated packages in SDK to create SBOMs (see https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/558)
* Successfully built core kit

**Terms of contribution:**


sbomtool/cmd/sbomtool/main.go
```diff
@@ -0,0 +1,180 @@
...
+	"strings"
+
+	"github.com/bottlerocket-os/twoliter/tools/sbomtool/go/pkg/generate"
+	"github.com/bottlerocket-os/twoliter/tools/sbomtool/go/pkg/logging"
+	"github.com/bottlerocket-os/twoliter/tools/sbomtool/go/pkg/merge"
+	"github.com/bottlerocket-os/twoliter/tools/sbomtool/go/pkg/validate"
+)
+
+// printHelp displays comprehensive usage information and command examples for sbomtool.
+func printHelp() {
```
ginglis13: consider using a multiline string here:

```
const usage = `
Usage:
    sbomtool [global options] command [command options]
Global Options:
    --help Show this help message
...
```

https://go.dev/doc/effective_go#web_server has an example

Or, as @arnaldo2792 pointed out, you could consider using cobra - although that's usually more useful when a binary has multiple subcommands (not necessarily many options). It's worth considering with `generate` and `merge` subcommands implemented already, if you anticipate the set of subcommands growing

sbomtool/README.md
```diff
@@ -0,0 +1,98 @@
+# sbomtool
+
+A Software Bill of Materials (SBOM) generation tool for the Bottlerocket SDK.
+
+## Overview
+
+`sbomtool` is a command-line utility that generates standardized Software Bill of Materials (SBOM) files for software packages. It analyzes a build directory to identify all components and dependencies, then produces SBOM files in industry-standard formats.
```
arnaldo2792: nit: it was a tradition in Bottlerocket to keep each sentence as its own line to facilitate reviews.

sbomtool/cmd/sbomtool/main.go
```diff
@@ -0,0 +1,184 @@
...
+		if *helpFlag || subcommandPos == len(os.Args) {
+			printHelp()
+			return 0
+		}
+	} else {
+		// Parse only the global flags before the subcommand
+		globalFlags.Parse(os.Args[1:subcommandPos])
+	}
+
+	// Configure logging based on global flag
```
arnaldo2792: I think most of the comments are unnecessary. It is clear in the code what's happening. Rather than explaining the what, the comment should explain a why. e.g. in Bottlerocket, they are used to explain when something goes off-script like overriding some C flags while compiling.

sbomtool/cmd/sbomtool/main.go
```diff
@@ -0,0 +1,184 @@
...
+	// If no subcommand is provided or "help" is the subcommand
+	if subcommandPos == len(os.Args) || os.Args[subcommandPos] == "help" {
+		printHelp()
+		return 0
+	}
+
+	// Get the subcommand
+	subcommand := os.Args[subcommandPos]
+
+	switch subcommand {
```
arnaldo2792: I know `cobra` is a widely used CLI tool in Go, but I'll let the Go experts advise whether adding that to this CLI would be unnecessary.

sbomtool/cmd/sbomtool/main.go
```diff
@@ -0,0 +1,184 @@
...
+			fmt.Fprintf(os.Stderr, "Error parsing generate command: %v\n", err)
+			return 1
+		}
+
+		slog.Info("Starting sbomtool generate",
+			"name", *generateName,
+			"build_dir", *generateBuilDir,
+			"out_dir", *generateOutDir,
+			"spdx", *generateSpdx,
+			"cyclonedx", *generateCyclonedx,
```
arnaldo2792: Shouldn't this be `debug` given all the information that was passed? Also, another place to use a widely used logging library `logrus`. This is used by projects like containerd. The advantage being that you can provide `context` to the logs.

sbomtool/go.work.sum
```diff
@@ -0,0 +1,3 @@
+github.com/containerd/typeurl v1.0.2 h1:Chlt8zIieDbzQFzXzAeBEF92KhExuE4p9p92/QmY7aY=
```
arnaldo2792, line 159: Can you please document either in the commit message or in the PR description why you had to do this?

sbomtool/pkg/generate/generate.go
```diff
@@ -0,0 +1,136 @@
...
+	if closeErr := f.Close(); closeErr != nil {
+		slog.Error("Failed to close file", "operation", operation, "error", closeErr)
+	}
+}
+
+// createSpdxSbom creates an SPDX SBOM file
+func createSpdxSbom(sbom sbom.SBOM, name string, outDir string) error {
+	cfg := spdxjson.DefaultEncoderConfig()
+	cfg.Pretty = false
+	cfg.Version = "2.3"
```
arnaldo2792, line 29: The latest SPDX versions for SBOMs seems to be 3.0.0:

https://spdx.dev/use/specifications/

Could you please elaborate on why supporting only 2.3? Also, if we will support multiple versions at the time, an abstraction over the versions could be needed here.

sbomtool/pkg/generate/generate.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	return nil
+}
+
+// createCyclonedxSbom creates a CycloneDX SBOM file
+func createCyclonedxSbom(sbom sbom.SBOM, name string, outDir string) error {
+	cfg := cyclonedxjson.DefaultEncoderConfig()
+	cfg.Pretty = false
+	cfg.Version = "1.6"
```
arnaldo2792, line 54: Does this mean `1.6.x` or `1.6` only. Because there is a 1.6.1 release of the CycloneDX specification.

https://github.com/CycloneDX/specification/releases

sbomtool/pkg/generate/generate_refactor.md
```diff
@@ -0,0 +1,213 @@
+# Proposed Refactoring for Better Testability
```
arnaldo2792: You may want to remove this file in this commit, it was removed later on the series.

arnaldo2792 (changes requested):

sbomtool/go.work.sum
rpkelly: This was because syft uses 3 different golang `xz` libraries and our license scanner wasn't happy with one of them.

sbomtool/pkg/generate/generate.go
rpkelly: We need to standardize on the version because each is a schema change (and therefore a different SBOM file) In this case, both Trivy and Syft support 2.3, but not 3.0

sbomtool/pkg/generate/generate.go
rpkelly: > Functionally equivalent to CycloneDX 1.6.0

The spec overview also calls out just `1.6` as the latest.

sbomtool/cmd/sbomtool/main.go
rpkelly: `logrus` predates `slog` but `slog` is now in the standard library, so waiting on golang experts to weigh in, it seems like a new project should prefer it.

This could definitely work as `debug`

Dockerfile
```diff
@@ -1286,3 +1320,5 @@ ENV GO_MAJOR="1.23"
 ENV NSS_DISABLE_PKIX_VERIFY=1
 
 CMD ["/bin/bash"]
+
+# =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
```
ginglis13: nit: extra cats?

Makefile
```diff
@@ -11,7 +11,13 @@ IMAGE_NAME ?= bottlerocket-sdk:$(VERSION)-$(SHORT_SHA)-$(DOCKER_ARCH)
 
 all: sdk
 
-sdk:
+# Clean up any previous sbomtool build artifacts
+clean-sbomtool:
+	@echo "Cleaning up previous sbomtool build artifacts..."
+	@rm -rf $(TOP)sbomtool/vendor $(TOP)sbomtool/sbomtool
+	@echo "Sbomtool artifacts cleaned."
```
ginglis13: Maybe we can use a `.dockerignore` instead

sbomtool/go.mod
```diff
@@ -0,0 +1,219 @@
...
+	google.golang.org/genproto v0.0.0-20241118233622-e639e219e697 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20241223144023-3abc09e42ca8 // indirect
+	google.golang.org/grpc v1.67.3 // indirect
+	google.golang.org/protobuf v1.36.4 // indirect
+	gopkg.in/warnings.v0 v0.1.2 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+// Replace xi2/xz with therootcompany/xz
+replace github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 => github.com/therootcompany/xz v1.0.0
```
ginglis13: This should be covered in `go.work`

sbomtool/README.md
```diff
@@ -0,0 +1,99 @@
...
+`sbomtool` is a command-line utility that generates standardized Software Bill of Materials (SBOM) files for software packages.
+ It analyzes a build directory to identify all components and dependencies, then produces SBOM files in industry-standard formats.
+
+## Features
+
+- Generate SBOM files in multiple formats:
+  - SPDX 2.3 (JSON)
+  - CycloneDX 1.6 (JSON)
+- Configurable logging levels
+- Thorough input validation
```
ginglis13: nit but I'd say this is a given not a feature

sbomtool/README.md
```diff
@@ -0,0 +1,99 @@
...
+- Future support for merging multiple SBOM files
```
ginglis13, line 15: I'm not through reading the code yet to see what's remaining for this support but maybe we should track this with an issue and point to it here

sbomtool/cmd/sbomtool/main.go
```diff
@@ -0,0 +1,180 @@
...
+			"log_level", *logLevelFlag)
+
+		// Validate all generate arguments
+		if err := validate.ValidateGenerateArgs(*generateName, *generateBuilDir, *generateOutDir, *generateSpdx, *generateCyclonedx); err != nil {
+			slog.Error("Invalid generate arguments", "error", err)
+			fmt.Fprintf(os.Stderr, "Invalid arguments: %v\n", err)
```
ginglis13: This applies to the whole PR - any reason why we're printing to both stderr and logging at an error level? Won't this defeat the purpose of configuring a log level?

sbomtool/cmd/sbomtool/main.go
```diff
@@ -0,0 +1,166 @@
...
+		"name", name,
+		"build_dir", buildDir,
+		"out_dir", outDir,
+		"spdx", spdx,
+		"cyclonedx", cyclonedx)
+
+	success, err := generate.Generate(name, spdx, cyclonedx, buildDir, outDir)
+	if err != nil {
+		slog.Error("SBOM generation failed", "error", err)
+		return fmt.Errorf("SBOM generation failed: %w", err)
```
ginglis13: This will log the same error twice

sbomtool/cmd/sbomtool/main.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if !success {
+		slog.Warn("No SBOM formats were generated")
+		return fmt.Errorf("no SBOM formats were generated")
```
ginglis13: nit: inconsistent capitalization 

Also, if log level is error, there will be both a duplicate warn and error log

sbomtool/pkg/generate/generate.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if !cyclonedx && !spdx {
+		slog.Warn("No SBOM format selected")
+		return false, nil
+	}
+
+	slog.Debug("Creating output directory", "path", outDir)
+	if err := os.MkdirAll(outDir, 0755); err != nil {
+		slog.Error("Failed to create output directory", "path", outDir, "error", err)
+		return false, fmt.Errorf("failed to create output directory: %w", err)
```
ginglis13: This log will bubble up to main, so you'll have duplicate logs

sbomtool/pkg/validate/validate.go
```diff
@@ -0,0 +1,151 @@
...
+
+	if level > 10 {
+		return fmt.Errorf("merge level too high (max 10)")
+	}
+
+	return nil
+}
+
+// ValidateGenerateArgs validates all arguments for the generate command.
+func ValidateGenerateArgs(name, buildDir, outDir string, spdx, cyclonedx bool) error {
```
ginglis13: This method isn't currently being used anywhere, maybe you meant for it to be used in the main command parsing for `sbomtool`? A similar function that does nearly the same validation exists there as `validateGenerateFlags`

rpkelly: Ah, this should have been removed when migrating to cobra.

sbomtool/go.mod
```diff
@@ -0,0 +1,216 @@
+module github.com/bottlerocket-os/twoliter/tools/sbomtool/go
```
ginglis13: Should this be the sdk?

```suggestion
module github.com/bottlerocket-os/bottlerocket-sdk/sbomtool
```

sbomtool/pkg/generate/generate.go
```diff
@@ -0,0 +1,130 @@
...
+	cfg := spdxjson.DefaultEncoderConfig()
+	cfg.Pretty = false
+	cfg.Version = "2.3"
+
+	encoder, err := spdxjson.NewFormatEncoderWithConfig(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create SPDX encoder: %w", err)
+	}
+
+	outputPath := filepath.Join(outDir, name+"-spdx.json")
```
ginglis13, line 36: definitely a nit but you could consider hoisting this and the cyclonedx suffix into constants in this package

ginglis13 (approved):

sbomtool/README.md
```diff
@@ -0,0 +1,97 @@
...
+- `--spdx`: Generate an SPDX SBOM
+- `--cyclonedx`: Generate a CycloneDX SBOM
+
+#### Merge (Future Feature)
+
+Merge multiple SBOM files:
+
+```
+sbomtool merge [options] file1 file2 [file3...]
+```
```
arnaldo2792, line 55: If it hasn't been implemented, why call it out?

sbomtool/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Output
+
+The tool generates SBOM files in the specified output directory:
+- `{name}-spdx.json`: SPDX format SBOM
+- `{name}-cyclonedx.json`: CycloneDX format SBOM
+
+## Implementation Details
+
+`sbomtool` uses the [Anchore Syft](https://github.com/anchore/syft) library for SBOM generation, which provides comprehensive package detection across various ecosystems.
```
arnaldo2792, line 87: nit: I prefer (and in the main Bottlerocket also used to prefer) the "citation" syntax:

```
sbomtool` uses the [Anchore Syft]

[Anchore Syft]: https://github.com/anchore/syft
```

sbomtool/cmd/sbomtool/main.go
```diff
@@ -0,0 +1,163 @@
...
+// createRootCommand creates and configures the root Cobra command for sbomtool.
+func createRootCommand() *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:   "sbomtool",
+		Short: "Software Bill of Materials (SBOM) generation tool",
+		Long: `sbomtool is a command-line utility that generates standardized Software Bill of Materials (SBOM) files for software packages.
+
+It analyzes a build directory to identify all components and dependencies, then produces SBOM files in industry-standard formats including SPDX 2.3 and CycloneDX 1.6.
+
+The tool is designed to integrate into build pipelines and supports comprehensive package detection across various ecosystems using the Anchore Syft library.`,
```
arnaldo2792: I wouldn't call out that the tool is designed to integrate into build pipelines, because it can only be used through the Bottlerocket SDK (unless people build it from source, in which case they could, but I don't think the tools in the SDK are meant to be used outside the context of the SDK).
```suggestion
```

sbomtool/cmd/sbomtool/main.go
```diff
@@ -0,0 +1,163 @@
...
+
+	generateCmd.MarkFlagRequired("name")
+	generateCmd.MarkFlagRequired("build-dir")
+	generateCmd.MarkFlagRequired("out-dir")
+	generateCmd.MarkFlagsOneRequired("spdx", "cyclonedx")
+
+	return generateCmd
+}
+
+// validateGenerateFlags performs comprehensive validation of generate command flags.
```
arnaldo2792:
```suggestion
validateGenerateFlags performs validation of generate command flags.
```

sbomtool/go.mod
```diff
@@ -0,0 +1,216 @@
+module github.com/bottlerocket-os/bottlerocket-sdk/sbomtool/go
+
+go 1.24.1
+
+toolchain go1.24.2
```
arnaldo2792: I'm not a go person, but shouldn't this match the version of the SDK?

sbomtool/go.work
```diff
@@ -0,0 +1,5 @@
+go 1.24.1
+
+use .
```
arnaldo2792: Same question as elsewhere about the Go version of the SDK.

sbomtool/pkg/generate/generate.go
```diff
@@ -0,0 +1,130 @@
...
+		"build_dir", buildDir,
+		"out_dir", outDir)
+
+	if !cyclonedx && !spdx {
+		slog.Warn("No SBOM format selected")
+		return false, nil
+	}
+
+	slog.Debug("Creating output directory", "path", outDir)
+	if err := os.MkdirAll(outDir, 0755); err != nil {
```
arnaldo2792, line 90: Why not use the same permissions like in the target files?

sbomtool/pkg/logging/logging.go
```diff
@@ -0,0 +1,39 @@
...
+	case "error":
+		slogLevel = slog.LevelError
+	default:
+		slogLevel = slog.LevelInfo // Default to info for unknown levels
+	}
+
+	// Create a JSON handler for structured logging
+	handler := slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
+		Level: slogLevel,
+	})
```
arnaldo2792, line 35: Does this mean that the logs will all output JSON instead of just plain log lines? If so, I'll prefer the plain text lines instead of JSON lines otherwise the logs in the builds will be hard to read (more than they are now)

sbomtool/pkg/validate/validate.go
```diff
@@ -0,0 +1,79 @@
...
+// This validation is intentionally permissive to support diverse package ecosystems
+// (RPM, Go, Rust, NPM, etc.) while preventing obvious security and filesystem issues.
+func ValidatePackageName(name string) error {
+	if name == "" {
+		return fmt.Errorf("package name cannot be empty")
+	}
+
+	if len(name) > 255 {
+		return fmt.Errorf("package name too long (max 255 characters)")
+	}
```
arnaldo2792, line 26: Is this validation a restriction in either of the specs?

sbomtool/pkg/validate/validate.go
```diff
@@ -0,0 +1,79 @@
...
+		// Disallow characters that could cause issues in file systems or shells
+		if r == '\x00' || r == '\n' || r == '\r' || r == '\t' {
+			return fmt.Errorf("package name contains invalid character at position %d", i)
+		}
+	}
+
+	// Prevent names that are just whitespace
+	if strings.TrimSpace(name) == "" {
+		return fmt.Errorf("package name cannot be only whitespace")
+	}
```
arnaldo2792: nit: You could move this after the length check to save you the empty `for` execution.

Dockerfile
```diff
@@ -823,6 +823,36 @@ RUN \
...
+USER builder
+WORKDIR /home/builder/sbomtool
+
+COPY ./sbomtool /home/builder/sbomtool
+
+# Copy license files
+COPY --chown=0:0 --from=sdk-rust-tools /usr/libexec/tools/ /usr/libexec/tools/
+RUN \
+  cp -p LICENSE-* /usr/share/licenses/sbomtool/ && \
+  # Use go work vendor instead of go mod vendor
```
arnaldo2792, line 841: The comment should explain why is `go mod work` used rather than `go mod vendor`.

Dockerfile
```diff
@@ -823,6 +823,36 @@ RUN \
...
+RUN \
+  mkdir -p /usr/libexec/tools /usr/share/licenses/sbomtool && \
+  chown -R builder:builder /usr/libexec/tools /usr/share/licenses/sbomtool
+
+USER builder
+WORKDIR /home/builder/sbomtool
+
+COPY ./sbomtool /home/builder/sbomtool
+
+# Copy license files
```
arnaldo2792: Shouldn't this be "Copy license scan files"?

macros/shared
```diff
@@ -294,6 +294,18 @@ CROSS_CMAKE_TOOLCHAIN_EOF\
...
+%_cross_sbom_package_dir %{_cross_sbom_dir}/%{_uncross_name}
+
+%cross_generate_sbom \
+  mkdir -p %{_builddir}/sbom-temp \
+  sbomtool generate --name %{name} --out-dir %{_builddir}/sbom-temp --build-dir %{_builddir} --spdx --cyclonedx
+
+%cross_install_sbom \
+  install -d %{buildroot}%{_cross_sbom_package_dir} \
+  install -p -m 0644 %{_builddir}/sbom-temp/%{name}-spdx.json %{buildroot}%{_cross_sbom_package_dir}/ \
+  install -p -m 0644 %{_builddir}/sbom-temp/%{name}-cyclonedx.json %{buildroot}%{_cross_sbom_package_dir}/
```
arnaldo2792, line 307: Why not just one single macro that generates the SBOM and installs it, like the `cross_attribution` macro? Also, shouldn't you add the sbom generation to the post-build:

https://github.com/bottlerocket-os/bottlerocket-sdk/blob/36019b16ad11b02914148a741dacb48a0e69455f/macros/shared#L342C2-L342C21

sbomtool/README.md
rpkelly: It currently exists in the cli, it just isn't implemented yet. This offers an explanation.

sbomtool/README.md
rpkelly: That is not the pattern used in Bottlerocket's README.md https://github.com/bottlerocket-os/bottlerocket/blob/develop/README.md?plain=1

sbomtool/cmd/sbomtool/main.go
rpkelly: Changed

sbomtool/pkg/validate/validate.go
rpkelly: It's based on filesystem. Most but not all impose that length limit.

macros/shared
rpkelly: I have a commit staged that adds it to post build. This one was just enough to get me to be able to test it by manually adding it to the packages.

rpkelly: And this can't be the same macro, because the generate needs to occur during the `%build` step, unlike the license scan which occurs in the `%install` step

sbomtool/go.mod
rpkelly: It will build, and this keeps it buildable on Amazon Linux for local testing. I can bump it to 1.24.2 though.

Dockerfile
rpkelly: I don't believe so? It's copying the LICENSE files as well as the license files from the vendor directory to `/usr/share/licenses` I can get rid of the comment though, since we do that same operation elsewhere and don't leave a comment

sbomtool/README.md
arnaldo2792: I guess it was used in releases.

arnaldo2792 (approved):

PR #306, opened 2025-11-05
add dependencies for libbpf and systemd-boot

bcressey:
**Issue number:**
N/A

**Description of changes:**
Add libbpf-devel to support downstream builds that use libbpf-rs.

Vendor clang and LLVM from the Rust build, to tighten control over the tool chain used to compile BPF programs.

Include pyelftools to enable building systemd-boot on aarch64.

**Testing done:**
Built a Rust program that used libbpf-rs to load a BPF LSM hook.

Built systemd-boot for arm64 on arm64. There are other problems with building for x86_64, but this is a start to unblock experiments.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -989,6 +989,7 @@ RUN \
     groff \
     kpartx \
     less \
+    libbpf-devel \
     libcap-devel \
     libkcapi-hmaccalc \
     lz4 \
--- a/Dockerfile
+++ b/Dockerfile
@@ -280,8 +280,8 @@ FROM sdk-libc AS sdk-rust
 
 USER root
 RUN \
-  mkdir -p /usr/libexec/rust && \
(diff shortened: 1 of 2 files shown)


piyush-jena (approved):

KCSesh:
nit: 

>Include pyelftools to enable building systemd-boot on aarch64.

Might be good information to include in the commit message as well that this is for aarch64

KCSesh (approved):

bcressey:
> Might be good information to include in the commit message as well that this is for aarch64

`pyelftools` is needed for both architectures. For x86_64, at first glance it seems like we need a 32-bit capable toolchain, but I haven't dug too deeply and possibly there's a workaround.

arnaldo2792 (approved):

PR #315, opened 2025-12-22
go: use runtime FIPS instead of compile-time boringcrypto

sky1122:
**Issue number:** #318 

**Description of changes:**

This is the minimal change for switching `boringcrypto` to runtime `GODEBUG=fips140=only`. Will remove the gofips build in a separate PR and release. 

FIPS compliance is now controlled at runtime via GODEBUG=fips140=only environment variable rather than compile-time GOEXPERIMENT=boringcrypto. 

**Testing done:**
Testing with [PR1](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/783), below showing the `gofips` point to regular go build.
original:
```
bash-5.1# grep -ao "GOEXPERIMENT=boringcrypto" /usr/fips/bin/aws-iam-authenticator
GOEXPERIMENT=boringcrypto
GOEXPERIMENT=boringcrypto
GOEXPERIMENT=boringcrypto
GOEXPERIMENT=boringcrypto
bash-5.1# exit
exit
[root@admin]# apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "2f4223e5",
    "pretty_name": "Bottlerocket OS 1.45.0 (aws-k8s-1.33-fips)",
    "variant_id": "aws-k8s-1.33-fips",
    "version_id": "1.45.0"
  }
}
```
Build with this change:
```
bash-5.2# grep -ao "GOEXPERIMENT=boringcrypto" /usr/fips/bin/kubelet
bash-5.2# grep -ao "GOEXPERIMENT=boringcrypto" /usr/fips/bin/aws-iam-authenticator
bash-5.2# exit
exit
[root@admin]# apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "55d04a85-dirty",
    "pretty_name": "Bottlerocket OS 1.53.0 (aws-k8s-1.33-fips)",
    "variant_id": "aws-k8s-1.33-fips",
    "version_id": "1.53.0"
  }
}
```
Testing with [PR1](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/783), and showing the fips work with go runtime variale
- Build bootstrap container that generates TLS certs on the node
- Configure bootstrap container on Bottlerocket node
- Deploy two registries with different cipher configurations
- Test image pulls

detail in: https://gist.github.com/sky1122/2fe6925332e7d9443eb149d5f45df116
result show the pull from non-fips cypher registry fail
- successful fips registry pull
```
Events:
  Type    Reason     Age   From               Message
  ----    ------     ----  ----               -------
  Normal  Scheduled  9s    default-scheduler  Successfully assigned default/test-pull-fips to ip-192-168-74-83.us-west-2.compute.internal
  Normal  Pulling    9s    kubelet            Pulling image "localhost:5000/test:latest"
  Normal  Pulled     9s    kubelet            Successfully pulled image "localhost:5000/test:latest" in 71ms (71ms including waiting). Image size: 3861737 bytes.
  Normal  Created    9s    kubelet            Created container: test
  Normal  Started    9s    kubelet            Started container test
```
- non-fips cypher registry pull
```
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type     Reason     Age                From               Message
  ----     ------     ----               ----               -------
  Normal   Scheduled  34s                default-scheduler  Successfully assigned default/test-pull-nonfips to ip-192-168-74-83.us-west-2.compute.internal
  Normal   Pulling    20s (x2 over 34s)  kubelet            Pulling image "localhost:5001/test:latest"
  Warning  Failed     20s (x2 over 34s)  kubelet            Failed to pull image "localhost:5001/test:latest": failed to pull and unpack image "localhost:5001/test:latest": failed to resolve image: failed to do request: Head "https://localhost:5001/v2/test/manifests/latest": remote error: tls: handshake failure
  Warning  Failed     20s (x2 over 34s)  kubelet            Error: ErrImagePull
  Normal   BackOff    7s (x2 over 33s)   kubelet            Back-off pulling image "localhost:5001/test:latest"
  Warning  Failed     7s (x2 over 33s)   kubelet            Error: ImagePullBackOff
```

**Terms of contribution:**


Dockerfile
```diff
@@ -641,21 +641,7 @@ RUN ./build-aws-lc.sh --arch="${ARCH}" --target="${TARGET}" --go-dir="${HOME}/sd
...
-  /home/builder/sdk-go/src/crypto/internal/boring/syso/goboringcrypto_linux_arm64.syso
-
-COPY --from=sdk-go-1.25-aws-lc-musl-x86_64 \
-  /home/builder/aws-lc/build/goboringcrypto_linux_amd64.syso \
-  /home/builder/sdk-go/src/crypto/internal/boring/syso/goboringcrypto_linux_musl_amd64.syso
-
-COPY --from=sdk-go-1.25-aws-lc-musl-aarch64 \
-  /home/builder/aws-lc/build/goboringcrypto_linux_arm64.syso \
-  /home/builder/sdk-go/src/crypto/internal/boring/syso/goboringcrypto_linux_musl_arm64.syso
+# FIPS/boringcrypto disabled - skipping aws-lc syso files
```
bcressey: These comments aren't necessary as they refer to code that no longer exists.

Dockerfile
```diff
@@ -1174,16 +1146,12 @@ COPY --chown=0:0 --from=sdk-go-1.24 \
   /home/builder/sdk-go/licenses/ \
   /usr/share/licenses/go-1.24/
 
-# Create Go trees for the different glibc and musl builds of the AWS-LC syso.
+# Create Go trees for musl builds (no boringcrypto syso files).
```
bcressey: The different glibc/musl paths shouldn't be necessary any longer. They were added because the syso files end up linked against a libc and consequently have different ABIs, meaning that the Go builds in turn needed to target one or the other.

Dockerfile
bcressey: This part of the change is incomplete - you're still building AWS-LC and patching the Go sources, you're just not including the syso in the SDK any longer.

You shouldn't end up building AWS-LC after this change.

sky1122: make sense since AWS-LC is no longer used. Will remove all the aws-lc related hack.

patches/go-1.25/0002-Always-restrict-boringcrypto-crypto-tls-to-FIPS.patch
```diff
@@ -1,60 +0,0 @@
-From 6e231b2606e477c823874c54bb12f9edf47dbaa2 Mon Sep 17 00:00:00 2001
```
bcressey, line 1: Other Go 1.24 and 1.25 patches should be renumbered so we continue to start with patch 0001 for each.

wrappers/go/gofips
```diff
@@ -10,10 +10,4 @@ if [ -n "${GO_VERSION:-}" ] ; then
   GO_MAJOR="${GO_VERSION%.*}"
 fi
 
-# prefer the musl flavor if GOLIBC indicates that we should
-if [ "${GOLIBC:-}" = "musl" ] ; then
-  GO_MAJOR+="-musl"
-fi
```
ginglis13, line 16: Should this also be removed from the other wrappers too? https://github.com/bottlerocket-os/bottlerocket-sdk/blob/1a788d2daeac7cc42a9d9a5aa4f3980ad1d7be4b/wrappers/go/go-latest#L15-L18

https://github.com/bottlerocket-os/bottlerocket-sdk/blob/1a788d2daeac7cc42a9d9a5aa4f3980ad1d7be4b/wrappers/go/gofmt#L13-L16

What about the `set_cross_go_flags_static` macro, which would with this change will no longer have the same effect?

https://github.com/bottlerocket-os/bottlerocket-sdk/blob/1a788d2daeac7cc42a9d9a5aa4f3980ad1d7be4b/macros/shared#L275-L276

I see we use that macro in building statically linked binaries for the amazon-ssm-agent: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/529e0b10c4d05fd6670f51fa91d1cb361537a4fd/packages/amazon-ssm-agent/amazon-ssm-agent.spec#L91

wrappers/go/go
```diff
@@ -10,9 +10,4 @@ if [ -n "${GO_VERSION:-}" ] ; then
   GO_MAJOR="${GO_VERSION%.*}"
 fi
 
-# prefer the musl flavor if GOLIBC indicates that we should
-if [ "${GOLIBC:-}" = "musl" ] ; then
-  GO_MAJOR+="-musl"
-fi
```
bcressey: This will cause the `%set_cross_go_flags_static` macro to stop working as intended.

You can see if redefining that macro to use glibc instead works to build `amazon-ssm-agent`, which is currently the only known user of that macro.

```
%set_cross_go_flags_static \
  %set_cross_go_env %{_cross_target} gnu _cross_go_extldflags_static
```

Ideally you would clean up `%set_cross_go_env` to not expect the second parameter (musl vs gnu) and then drop the GOLIBC export.

wrappers/go/gofips
sky1122: Thanks for calling out! 
I will remove the musl logic from bottlerocket-sdk/wrappers/go/go-latest and bottlerocket-sdk/wrappers/go/gofmt.

For `set_cross_go_flags_static`, I will like to keep it unchanged to minimize spec file changes in core-kit. The macro still sets GOLIBC=musl, but since I removed the GOLIBC check from the Go wrappers, this environment variable is now ignored by Go. It still serves its purpose for C compiler selection (musl-gcc), just no longer affects Go tree selection.

Dockerfile
```diff
@@ -365,8 +365,7 @@ COPY --from=sdk-rust /usr/libexec/llvm/ /usr/
 
 # =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
-# Merge the LLVM install with the rest of our toolchains and C libraries for
-# later use by the AWS-LC builds.
+# Merge the LLVM install with the rest of our toolchains and C libraries.
```
bcressey: You can remove this whole `sdk-libc-llvm` stage and have Go depend on `sdk-libc` instead, since we no longer need LLVM / clang during the Go build.

wrappers/go/gofips
ginglis13: Ah, I'm seeing `-static-pie` is set and `x86_64-bottlerocket-linux-musl-gcc` is set regardless of GOLIBC when I use the ` %set_cross_go_flags_static` macro, so amazon-ssm-agent will continue to be statically compiled

bcressey: I'd still prefer to update `%set_cross_go_env` not to take the second parameter, which is used to set `GOLIBC`, and then update the two macros that call it not to pass the second parameter.

sky1122:
commit to father clean up the prepare stage and macro.

patches/aws-lc/0001-Avoid-weak-symbols-when-building-for-Go.patch
```diff
@@ -1,34 +0,0 @@
-From 0fb9b17aa5d0d63d17e6914f7c34358db69e5647 Mon Sep 17 00:00:00 2001
```
bcressey, line 1: typo in commit message: "delete aws-lc patche and renumbers"

bcressey (approved):

sky1122:
pushed to fix typo

ginglis13 (approved):

Issue #316, opened 2025-12-23
SBOM merge should include root nodes


rpkelly:
**Image I'm using:**
N/A

**Issue or Feature Request:**
On SBOMs where there are no dependencies and only 1 root file, the merge function seems to miss that root file when creating the AMI sbom. This file should be included.

PR #317, opened 2025-12-30
fix: sbomtool merge losing packages and properties

rpkelly:
Fix three bugs in the SBOM merge process:

1. Remove DocumentRoot filter that incorrectly dropped primary packages
   - The filter was removing packages whose ID contained 'DocumentRoot'
   - CycloneDX puts the scan target in metadata.component, not components[]
   - When syft decodes this, it creates a package with DocumentRoot in ID
   - Now we only filter packages with empty names

2. Preserve Locations and Metadata during package merging
   - Locations (file paths) were not being merged, causing all syft:location:* properties to be lost
   - Metadata was not being preserved, causing all syft:metadata:* and syft:package:* properties to be lost
   - Now we merge all Locations from duplicate packages (union)
   - Now we preserve Metadata from first package with non-nil metadata

3. Use CycloneDX 1.6 for output encoding
   - The default encoder used CycloneDX 1.2 which doesn't support the properties field on components (added in 1.3)
   - Now explicitly use CycloneDX 1.6 encoder to preserve all properties

**Issue number:** #316 

**Description of changes:**
Changed SBOM tool to preserve more information during a merge. Changes now keep the document root package, as well as ensure that CycloneDX encoding is 1.6 for more metadata.

**Testing done:**
Generated SBOMs for Bottlerocket core and kernel kits. See [here](https://gist.github.com/rpkelly/a2e6e5498377afa173f00bab8cb6931e) for full before and partial after (full after seems to give gist problems)

**Terms of contribution:**


cezar-r (approved):

sbomtool/internal/processor/processor.go
```diff
@@ -1,5 +1,3 @@
-// Package processor provides Syft-based SBOM processing capabilities optimized for Bottlerocket builds.
-// It configures Syft's catalogers for comprehensive package detection including Go and Rust binary analysis.
```
ginglis13, line 2: Looks like a few docstrings in this file got removed

ginglis13 (commented):
in "Testing Done", could you also share some before/after SBOM examples for a variant?

ginglis13 (approved):

cezar-r (approved):

PR #328, opened 2026-02-03
macros: add configurable SBOM build directory

rpkelly:
**Description of changes:**
Rust packages i n the core and kernel kit redefine the generate sbom macro purely so that that they can change the build_dir used by the macro. By exposing the directory as a different, redefinable macro we simplify what these sorts of packages need to change.

Removing the `buildsubdir` broadens the scope of packages that will generate an SBOM package. While many of these will be blank SBOMs, some will not so, we'll opt for the widest possible net.

**Testing done:**
Built an SDK and then a core kit

**Terms of contribution:**

### DIFF
--- a/macros/shared
+++ b/macros/shared
@@ -332,9 +332,11 @@ This package provides SBOM files for %{name}.\
 %[ 0%{?_enable_sbom_packages} > 0 ? "%{?buildsubdir:%{expand:%%global __sbom_package 1}%(echo "%{_sbom_template}" > %{specpartsdir}/rpm-sbom.specpart)}" : "" ]\
 %{nil}
 
+%_cross_sbom_build_dir %{_builddir}
+
 %cross_generate_sbom \
   mkdir -p %{_builddir}/sbom-temp \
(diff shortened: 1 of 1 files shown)


bcressey:
> Removing the buildsubdir broadens the scope of packages that will generate an SBOM package. While many of these will be blank SBOMs, some will not so, we'll opt for the widest possible net.

Can you say more about this? I don't really follow the downstream implications.

rpkelly:
>Can you say more about this? I don't really follow the downstream implications.

@bcressey For builders this will mean that some packages, especially ones that only container pre-built binaries, or ones that are only compose of configuration files, will now generate a `package-sbom` rpm. In most cases this will lead to a pretty bland sbom (assuming most cases are the latter case of just configuration files) since there's not much to put in that sort of sbom. So the result will be a small rpm with an sbom that mostly just says the package exists.

This won't change much during overall SBOM generation for images since we're already using the RPMs themselves to generate part of the SBOM, but it may lead to more complete information in some cases (and prevents cases where you have 100 rpms but only 97 matching SBOM rpms)

bcressey (approved):

ginglis13 (approved):

PR #362, opened 2026-05-19
license-scan: update cargo dependencies

jpculp:
**Description of changes:**

Updated `license-scan` dependencies to current versions and modernized the codebase. Replaced `structopt` with `clap`, removed `lazy_static` in favor of `std::sync::LazyLock`, and bumped `askalono`, `cargo_metadata`, `spdx`, and `toml` to their latest major versions. Also bumped the SPDX license list data from **3.26.0** to **3.28.0**.

`deny.toml` was cleaned up to remove the `structopt` skip-tree and the now unused **BSL-1.0** allow entry.

`clarify.toml` was updated for `spdx` **0.13** which changed its license from `MIT OR Apache-2.0` to `Apache-2.0` only.

**Testing done:**

Built old (develop) and new scanner binaries locally. Ran both against `cargo-deny` **0.18.1** and `govmomi` **0.46.3** (the same targets scanned in the Dockerfile). Old used `SPDX` **3.26.0** data, new used **3.28.0**. Output attribution files are identical between old and new for both targets.

**Terms of contribution:**

### DIFF
--- a/license-scan/deny.toml
+++ b/license-scan/deny.toml
@@ -5,7 +5,7 @@ confidence-threshold = 0.93
 
 allow = [
     "Apache-2.0",
-    "BSL-1.0",
+    # "BSL-1.0",
     "MIT",
     "Unlicense",
     "Unicode-3.0",
@@ -19,6 +19,14 @@ exceptions = [
 # Deny multiple versions or wildcard dependencies.
 multiple-versions = "deny"
 wildcards = "deny"
+
+skip = [
+    # transient: getrandom 0.2 from twox-hash (removed when default features disabled)
+    # and 0.3 from cc build-dep (removed when askalono bumps to 0.5)
+    { name = "getrandom" },
+    # transient: tempfile 3.15 pins rustix 0.38 which needs windows-sys 0.59
+    { name = "windows-sys" },
+]
 
 skip-tree = [
     # structopt is using an older syn
--- a/license-scan/deny.toml
+++ b/license-scan/deny.toml
@@ -28,10 +28,6 @@ skip = [
     { name = "windows-sys" },
 ]
 
-skip-tree = [
-    # structopt is using an older syn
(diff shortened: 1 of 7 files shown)


jpculp:
Ran `cargo fmt` and tweaked `clarify.toml`.

piyush-jena (approved):

license-scan/deny.toml
```diff
@@ -5,7 +5,7 @@ confidence-threshold = 0.93
 
 allow = [
     "Apache-2.0",
-    "BSL-1.0",
+    # "BSL-1.0",
```
arnaldo2792, line 8: If this is not needed anymore, why keep it as a comment?

license-scan/src/main.rs
```diff
@@ -19,28 +20,27 @@ use std::fmt;
 use std::fs;
 use std::hash::Hasher;
 use std::path::{Path, PathBuf};
-use structopt::StructOpt;
 use walkdir::WalkDir;
 
-#[derive(Debug, StructOpt)]
+#[derive(Debug, Parser)]
```
arnaldo2792, line 25: Was this change needed because `structopt` is unmaintained? Could you please document that in either the commit message or the PR description?.

arnaldo2792 (approved):

license-scan/src/main.rs
jpculp: I think `structopt` being unmaintained has more or less been common knowledge since 2020. We ported over the rest of Bottlerocket to `clap` and `argh` several years ago so I was surprised to find it here.

license-scan/deny.toml
jpculp: In other parts of bottlerocket we comment out approved, but unused, licenses to reduce confusion when it comes time to add them back in. If anything we should probably be restoring the licenses that have been removed up to this point (BSD-2-Clause, BSD-3-Clause, etc.)

